PR #1, opened 2022-10-26
Initial commit.

sergey-borovkov:
Login is implemented. Token is verified as working when making a request to get a non existing group. There are tests that check that response for correct and incorrect token is handled. In case of correct token path $HOME/.screenly is checked to have the token stored upon successful login.


src/commands.rs
```diff
@@ -0,0 +1,21 @@
+use crate::Authentication;
+
+
+
+
```
renatgalimov, line 5: Could you clean the formatting, please?

src/authentication.rs
```diff
@@ -0,0 +1,146 @@
...
+        assert!(path.exists());
+        let contents = fs::read_to_string(path).unwrap();
+        assert!(contents.eq("correct_token"));
+        tmp_dir.close().unwrap();
+    }
+
+    #[test]
+    fn test_verify_and_store_token_when_token_is_invalid() {
+        let tmp_dir = TempDir::new("invalid").unwrap();
+        env::set_var("HOME", tmp_dir.path().to_str().unwrap());
```
renatgalimov, line 130: This won't work. Tests run in parallel but inside one process. Modifying the environment inside the test will change the environment for whole process and will affect all other test threads.

src/authentication.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::authentication::Config;
+    use crate::Authentication;
+    use httpmock::{Method::GET, MockServer};
+    use simple_logger::SimpleLogger;
+    
+
```
renatgalimov, line 95: Too many whitespaces. Could we use rustfmt?

renatgalimov (changes requested):

src/commands.rs
sergey-borovkov: done

src/authentication.rs
sergey-borovkov: done

renatgalimov (approved):

PR #16, opened 2022-12-09
Support web asset upload

sergey-borovkov:
Supports web asset upload. Also uses info! and error! for logging.

## What does this PR do?
Support urls in `asset add` command.
-
## How has this been tested?
Unit test and manual upload
## Checklist before merging

- [x] If have added tests.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -5,6 +5,7 @@ use humantime::format_duration;
 use indicatif::{ProgressBar, ProgressStyle};
 use std::fs::File;
 use std::time::Duration;
+use log::info;
 use thiserror::Error;
 
 use prettytable::{row, Table};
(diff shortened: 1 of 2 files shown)


renatgalimov (approved):

src/commands.rs
```diff
@@ -250,16 +251,40 @@ impl AssetCommand {
...
+    fn add_web_asset(&self, url: &str, headers: &HeaderMap, payload: &HashMap<&str, &String>) -> anyhow::Result<Assets, CommandError> {
+        let response = self
+            .authentication
+            .build_client()?
+            .post(url)
+            .json(payload)
+            .headers(headers.clone())
+            .send()?;
+
+        if response.status().as_u16() != 201 {
```
renatgalimov, line 263: You should use https://docs.rs/reqwest/0.9.21/reqwest/struct.StatusCode.html

sergey-borovkov: what's the difference? I don't see it

src/commands.rs
```diff
@@ -250,16 +251,40 @@ impl AssetCommand {
...
+        Ok(Assets::new(serde_json::from_str(&response.text()?)?))
+    }
+
     pub fn add(&self, path: String, title: String) -> anyhow::Result<Assets, CommandError> {
         let url = format!("{}/v4/assets", &self.authentication.config.url);
 
         let mut headers = HeaderMap::new();
         headers.insert("Prefer", "return=representation".parse()?);
 
+        if path.starts_with("http") || path.starts_with("https") {
```
korvyashka, line 277: isn't path.starts_with("http") enough? path.starts_with("https") evaluated to true - means path.starts_with("http") also will be evaluated to true.

korvyashka (approved):

src/commands.rs
sergey-borovkov: true! though I guess I need to do http: and https: because local files can start with http as well.

PR #17, opened 2022-12-09
Adds workflow for release.

sergey-borovkov:
## What does this PR do?
Adds workflow for release.
-
## How has this been tested?
Not yet.
## Checklist before merging

- [ ] If have added tests.
Build and upload binaries on release action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,9 @@
 # .github/workflows/release.yml
 
 on:
-  release:
-    types: [created]
+  push:
+    tags:
+      - '*'
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
+# .github/workflows/release.yml
+
+on:
+  release:
+    types: [created]
```
vpetersson, line 5: We should probably do something like this instead: https://github.com/Screenly/Anthias/blob/master/.github/workflows/build-balena-disk-image.yaml#L2-L5

vpetersson (approved):

PR #36, opened 2022-12-28
New asset commands

sergey-borovkov:
## What does this PR do?
Supports `inject-js` and `set-headers` commands for web assets. 
Inject-js allows setting JS injection code for web assets. Injected code will be executed during playback on the device once the asset is loaded.

set-headers similarly works on web assets setting values of custom HTTP headers.
Usage is a bit awkward but I did not come up with better syntax:
```screenly asset set-headers 01FJVFP0QVD1E5TA70SA66AJPF --header k=v --header k2=v2```

Also adds `basic-auth` command that allows setting the basic auth header by accepting username and password.

T6482
## How has this been tested?
New unit tests + manual use.

## Checklist before merging

- [x] If have added tests.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -344,7 +344,7 @@ impl AssetCommand {
         Ok(Assets::new(serde_json::from_str(&response.text()?)?))
     }
 
-    pub fn set_headers(
+    pub fn set_web_asset_headers(
         &self,
         id: &str,
         headers: Vec<(String, String)>,
@@ -810,7 +810,9 @@ mod tests {
         let authentication = Authentication::new_with_config(config);
         let asset_command = AssetCommand::new(authentication);
         let headers = vec![("k".to_owned(), "v".to_owned())];
-        assert!(asset_command.set_headers("test-id", headers).is_ok());
+        assert!(asset_command
+            .set_web_asset_headers("test-id", headers)
+            .is_ok());
     }
 
     #[test]
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -326,8 +326,7 @@ impl AssetCommand {
 
 #[cfg(test)]
 mod tests {
-    use crate::authentication::Config;
-    use crate::{Authentication, Formatter, OutputType};
+    use super::*;
     use httpmock::{Method::GET, MockServer};
 
     use envtestkit::lock::lock_test;
@@ -336,7 +335,7 @@ mod tests {
     use std::ffi::OsString;
     use std::fs;
 
-    use crate::commands::{AssetCommand, Assets, ScreenCommand, Screens};
+    use crate::authentication::Config;
     use serde_json::{json, Value};
     use tempdir::TempDir;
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by vpetersson** [comment]
>  screenly asset set-headers 01FJVFP0QVD1E5TA70SA66AJPF "{\"header1\": \"value1\", \"header2\": \"value2\"}"

There's no reason why we need to have JSON style input. It would make more sense to set these like you would with say environment variables in Docker (e.g. `-h header1=value`).

**Review by renatgalimov — APPROVED** [review]

**Comment by sergey-borovkov** [comment]
> > screenly asset set-headers 01FJVFP0QVD1E5TA70SA66AJPF "{"header1": "value1", "header2": "value2"}"
> 
> There's no reason why we need to have JSON style input. It would make more sense to set these like you would with say environment variables in Docker (e.g. `-h header1=value`) and then just repeat that multiple times.

Done though without short `-h` - it's used for help already.

File: src/commands.rs [hunk 1]
```diff
@@ -318,6 +344,25 @@ impl AssetCommand {
         Ok(Assets::new(serde_json::from_str(&response.text()?)?))
     }
 
+    pub fn set_headers(
```
korvyashka, line 347: set_asset_headers may be - just to differentiate from actual request headers.

src/commands.rs
```diff
@@ -318,6 +344,25 @@ impl AssetCommand {
...
+        &self,
+        id: &str,
+        headers: Vec<(String, String)>,
+    ) -> anyhow::Result<(), CommandError> {
+        let endpoint = format!("v4/assets?id=eq.{}", id);
+        let map: HashMap<_, _> = headers.into_iter().collect();
+        patch(&self.authentication, &endpoint, &json!({ "headers": map }))
+    }
+
+    pub fn inject_js(&self, id: &str, js_code: &str) -> anyhow::Result<(), CommandError> {
```
korvyashka, line 357: I guess it is designed to be separate commands (set_headers and inject_js)- right?

src/commands.rs
```diff
@@ -346,23 +390,26 @@ mod tests {
...
...\",\"uptime\":\"230728.38\",\"load_avg\":\"0.14\",\"signal_strength\":null,\"interface\":\"eth0\",\"debug\":false,\"location\":\"Kamsko-Ust'inskiy rayon, Russia\",\"team\":\"016343c2-82b8-0000-a121-e30f1035875e\",\"timezone\":\"Europe/Moscow\",\"type\":\"hardware\",\"hostname\":\"srly-4shnfrdc5cd2p0p\",\"ws_open\":false,\"status\":\"Offline\",\"last_screenshot\":\"https://us-assets.screenlyapp.com/01CD1W50NR000A28F31W83B1TY/screenshots/01F98G8MJB6FC809MGGYTSWZNN/5267668e6db35498e61b83d4c702dbe8\",\"in_sync\":false,\"software_version\":\"Screenly 2 Player\",\"hardware_version\":\"Raspberry Pi 3B\",\"config\":{\"hdmi_mode\": 34, \"hdmi_boost\": 2, \"hdmi_drive\": 0, \"hdmi_group\": 0, \"verify_ssl\": true, \"audio_output\": \"hdmi\", \"hdmi_timings\": \"\", \"overscan_top\": 0, \"overscan_left\": 0, \"use_composite\": false, \"display_rotate\": 0, \"overscan_right\": 0, \"overscan_scale\": 0, \"overscan_bottom\": 0, \"disable_overscan\": 0, \"shuffle_playlist\": false, \"framebuffer_width\": 0, \"use_composite_pal\": false, \"framebuffer_height\": 0, \"hdmi_force_hotplug\": true, \"use_composite_ntsc\": false, \"hdmi_pixel_encoding\": 0, \"play_history_enabled\": false}}]").unwrap();
```
korvyashka, line 394: I guesd you could do raw string r#" to get rid of so much escaping

korvyashka (approved):

src/commands.rs
sergey-borovkov: yes

src/commands.rs
sergey-borovkov: Sure when I get to it. I tried using json!() everywhere instead but the macro hit recursion limit.

Issue #12, opened 2022-11-14
Change how tokens are inputed


vpetersson:
We currently enter tokens this way:

```bash 
$ screenly login [redacted]
Login credentials have been saved.
```

There issue with this is that by default (unless you prefix the command with a space), the token will be saved in your shell's history. This should be considered a security issue.

A better approach is to simply ask the user for this in an interactive dialogue.

sergey-borovkov:
Fixed in https://github.com/Screenly/cli/pull/44

PR #43, opened 2023-01-28
Ask for API token interactively to avoid having password in terminal …

sergey-borovkov:
…history.

## What does this PR do?
Users are now asked to input API token in the interactive dialogue. The API token is not saved to the terminal history or displayed in the terminal.
https://github.com/Screenly/cli/issues/12
## How has this been tested?
Manually
## Checklist before merging

- [ ] If have added tests.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,7 @@ use clap::{command, Parser, Subcommand};
 
 use http_auth_basic::Credentials;
 use log::{error, info};
+use rpassword::read_password;
 use simple_logger::SimpleLogger;
 use std::io::Write;
 use std::{fs, io};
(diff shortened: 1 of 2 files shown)


renatgalimov (approved):

src/main.rs
```diff
@@ -229,23 +230,28 @@ fn main() {
...
-                info!("Login credentials have been saved.");
-                std::process::exit(0);
-            }
-
-            Err(e) => match e {
-                AuthenticationError::WrongCredentials => {
-                    error!("Token verification failed.");
-                    std::process::exit(1);
+        Commands::Login {} => {
+            print!("Enter your API Token: ");
```
korvyashka, line 234: Could it be replaced by info!?

src/main.rs
```diff
@@ -229,23 +230,28 @@ fn main() {
...
+            std::io::stdout().flush().unwrap();
+            let token = read_password().unwrap();
```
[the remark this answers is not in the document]
sergey-borovkov: Its main what else can we do

src/main.rs
sergey-borovkov: No, because that will add a bunch of visual trash to the line that tasks to user

Issue #41, opened 2023-01-10
Tweak the way headers are set


vpetersson:
Currently, the way we set headers for web assets is as follows:

```bash
$ screenly asset $ASSET_ID set-headers —header k=v
[...]
```

When we do this, we wipe all existing headers in the process. This is somewhat unexpected.

An alternative approach is to add the following `headers` sub-commands:

* add
  * Adds on or more headers (passed on as comma separated key-value pairs)
* remove 
  * Removes on or more headers (passed on as comma separated key-value pairs)
* update
  * Updates on or more headers (passed on as comma separated key-value pairs)
* clear
  * Removes all headers

An alternative approach would be something like this:

```bash
$ screenly asset $ASSET_ID headers add k=v,foo=bar
[...]
```

If you then just want to update say the key `k`, you can use 

```bash
$ screenly asset $ASSET_ID headers update k=foobar
[...]
```

This is open for debate, but I think this would be more user friendly.

aljungberg:
Great write-up. Here's my take, two things.

First, not sure we should split this subcommand into two words (`headers add` versus `add-headers`). Barring any good reasons to deviate, I'd like to stay consistent with the `gcloud` command as a kind of best practice or "industry standard". And they don't split the operations. For example they have `add-tags` and `remove-tags`:

```
➜  ~ gcloud compute instances --help | rg tags
     add-tags
        Add tags to Compute Engine virtual machine instances.
     remove-tags
        Remove tags from Compute Engine virtual machine instances.
```

That said, allow me to immediately deviate from `gcloud` (they don't do what I'm about to propose, but neither do they do what Viktor proposed as you see above).

I think the use case of wanting to clobber all existing headers is valid and less surprising in some cases. Like if I have a cron job to update an asset with a new set of authorise headers or whatever every day due to security rotation, maybe something weird where even the header names change, it would be far easier to just set exactly what I want in one go, knowing all old headers are properly gone at the same time.

My set of commands would be like this:

- `set-headers [header1=value1[,header2=value2[,...]]]`: replaces all headers of the asset with the given headers (when an empty string is given, e.g. `--set-headers ""`, all existing headers are removed, if any)
- `update-headers <header1=value1[,header2=value2[,...]]>`: updates only the given headers (adding them if new), leaving any other headers unchanged

One elegant bit is that we don't need separate `add` or `clear` variants in this case:

- you can add with `update` 
- and clear with `set-headers` 

[...]

PR #44, opened 2023-01-29
Tweak the way headers are set

sergey-borovkov:
https://github.com/Screenly/cli/issues/41
## Checklist before merging

- [x] If have added tests.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 
 use humantime::format_duration;
 use indicatif::{ProgressBar, ProgressStyle};
-use log::info;
+use log::{debug, info};
 use std::fs::File;
 use std::time::Duration;
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


src/commands.rs
```diff
@@ -354,6 +354,36 @@ impl AssetCommand {
...
+        id: &str,
+        headers: Vec<(String, String)>,
+    ) -> anyhow::Result<(), CommandError> {
+        // if no headers provided update has nothing to do
+        if headers.is_empty() {
+            return Ok(());
+        }
+
+        let mut new_headers: HashMap<_, _> = headers.into_iter().collect();
+        let asset = self.get(id)?;
```
korvyashka, line 368: You can request headers only to reduce payload for get - /v4/assets?select=headers

src/commands.rs
```diff
@@ -354,6 +354,36 @@ impl AssetCommand {
...
+        if let Some(assets) = asset.value.as_array() {
+            if assets.is_empty() {
+                return Err(CommandError::MissingField);
+            }
+            let headers = assets[0].get("headers").ok_or(CommandError::MissingField)?;
+            let old_headers = serde_json::from_value::<HashMap<String, String>>(headers.clone())?;
+            debug!("Old headers {:?}", &old_headers);
+            for (key, value) in old_headers {
+                new_headers.entry(key).or_insert(value);
```
korvyashka, line 377: Hm - do you need to serialize/deserialize old_headers to/from json to mutate? It could be HashMap

korvyashka (changes requested):

src/commands.rs
sergey-borovkov: But it's the only place only headers are needed, and a single asset I don't see the reason to complicate it

src/commands.rs
sergey-borovkov: How else can I iterate over then ? I did not find a way if you know how please tell me

src/commands.rs
korvyashka: got it

src/commands.rs
korvyashka: HashMap supports iter method:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.iter

sergey-borovkov: What hashmap? It's serde Value

korvyashka (approved):

PR #47, opened 2023-02-13
Initial edge apps command support

sergey-borovkov:
## What does this PR do?
Adds support for 2 commands for edge apps:
Init - initializes skeleton manifest file for edge application in the specified directory. User is expected to fill out all fields right now.
Publish - command checks that manifest has all necessary fields filled out, publishes manifest to our store, retrieves its id, and other generated fields. Once done it updates the manifest skeleton field with retrieved values.
T6692.

## How has this been tested?
Manually and new unit tests.
## Checklist before merging

- [x] If have added tests.


src/cli.rs
```diff
@@ -274,189 +292,239 @@ pub fn handle_cli(cli: &Cli) {
...
+fn transform_edge_app_path_to_manifest(path: &Option<String>) -> PathBuf {
+    let mut p = "screenly.yml".to_owned();
+    if let Some(path_input) = path {
+        let mut buf = PathBuf::from(path_input);
+        buf.push("screenly.yml");
+        p = buf.to_str().unwrap_or("screenly.yml").to_owned();
+    }
+
+    PathBuf::from(p)
+}
```
renatgalimov, line 336:
```suggestion
fn transform_edge_app_path_to_manifest(path: &Option<String>) -> PathBuf {
    if let Some(path_input) = path {
        let mut buf = PathBuf::from(path_input);
        buf.push("screenly.yml");
        return buf;
    } 
    PathBuf::from("screenly.yml")
}
```

src/cli.rs
```diff
@@ -274,189 +292,239 @@ pub fn handle_cli(cli: &Cli) {
...
 
-                        if title != user_input.trim() {
-                            error!("The title you entered is incorrect. Aborting.");
+fn transform_edge_app_path_to_manifest(path: &Option<String>) -> PathBuf {
+    let mut p = "screenly.yml".to_owned();
+    if let Some(path_input) = path {
+        let mut buf = PathBuf::from(path_input);
+        buf.push("screenly.yml");
+        p = buf.to_str().unwrap_or("screenly.yml").to_owned();
```
renatgalimov, line 332:
```suggestion
        let mut buf = PathBuf::from(path_input);
        buf.push("screenly.yml");
        return buf;
```

src/cli.rs
```diff
@@ -274,189 +292,239 @@ pub fn handle_cli(cli: &Cli) {
...
+            let asset_command = commands::asset::AssetCommand::new(authentication);
+            handle_command_execution_result(asset_command.add(path, title), json);
+        }
+        AssetCommands::Delete { uuid } => {
+            let asset_command = commands::asset::AssetCommand::new(authentication);
+            match get_asset_title(uuid, &asset_command) {
+                Ok(title) => {
+                    info!("You are about to delete the asset named \"{}\".  This operation cannot be reversed.", title);
+                    info!("Enter the asset title to confirm the asset deletion: ");
+                    io::stdout().flush().unwrap();
```
renatgalimov, line 417: Do we support flags like `--force`, or `--noinput` to skip confirmation steps?

src/commands/edge_app.rs
```diff
@@ -0,0 +1,241 @@
...
+        edge_app_json.remove("created_at");
+        edge_app_json.remove("created_by");
+        edge_app_json.remove("updated_at");
+        edge_app_json.remove("permissions");
+
+        let mut payload = HashMap::<String, String>::new();
+        // for now all values are fields in the manifest are required to be non-empty.
+        // if we change that we will need to have a list of required non-empty values.
+        for (k, v) in edge_app_json {
+            debug!("{k}{v:?}");
```
renatgalimov, line 91: Add at least some information on what's going on when debugging.

```suggestion
        // if we change that we will need to have a list of required non-empty values.
        debug!("Edge app JSON:");
        for (k, v) in edge_app_json {
            debug!("{k}{v:?}");
```

src/commands/edge_app.rs
```diff
@@ -0,0 +1,241 @@
...
+            id: "01GS5H2CX6Y10ZRJHEDQPEWN4E".to_string(),
+            created_at: "2023-02-13T13:57:43+00:00".to_string(),
+            created_by: "01D0C6YJA8000A7JEN0230MWGF".to_string(),
+            updated_at: "2023-02-13T13:57:43.97356+00:00".to_string(),
+            permissions: vec![],
+
+            ..manifest.clone()
+        }];
+
+        let mut b = serde_json::to_value(&manifest).unwrap();
```
renatgalimov, line 195: Use normal variable names please.

src/commands/edge_app.rs
```diff
@@ -0,0 +1,241 @@
...
+
+            then.status(201)
+                .json_body(serde_json::to_value(&published_manifest).unwrap());
+        });
+
+        let config = Config::new(mock_server.base_url());
+        let authentication = Authentication::new_with_config(config);
+        let command = EdgeAppCommand::new(authentication);
+
+        let p = tmp_dir.path().join("screenly.yml");
```
renatgalimov, line 222: Use meaningful variable names please.

renatgalimov (changes requested):

src/cli.rs
sergey-borovkov: no

src/commands/edge_app.rs
sergey-borovkov: debug does exactly that though, no? It will show context and what line it comes from

src/commands/edge_app.rs
sergey-borovkov: why? it's test code. Letter b should be fine for variables whose whole purpose is to hold a temporary value (binding).

src/commands/edge_app.rs
sergey-borovkov: again - this is test code - why does "p" not work?

src/cli.rs
renatgalimov, line 336: I'd say that this would look more natural with combinators. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8b2253fedf2f34b607711f4d81bc853b
```suggestion
fn transform_edge_app_path_to_manifest(path: &Option<String>) -> PathBuf {
    let mut result = PathBuf::from(
        path_input.as_ref().map(String::as_str).unwrap_or("")
    );
    result.push("screenly.yml");
    result
}
```

renatgalimov (changes requested):

src/cli.rs
sergey-borovkov: (but maybe in the future we will add it).

renatgalimov (approved):

src/commands/edge_app.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    pub fn init(self, path: &Path) -> Result<(), CommandError> {
+        let mut object = serde_yaml::to_value(EdgeAppManifest::default())?;
+        let map = object.as_mapping_mut().ok_or(CommandError::MissingField)?;
+
+        // following fields will be generated server side when publishing.
+        map.remove("id");
+        map.remove("created_at");
+        map.remove("created_by");
+        map.remove("updated_at");
```
korvyashka, line 52: It seems you don't use these fields at all: created_at, created_by and updated_at both at init and on publish - may be just remove them from json struct - and they I guess should be ignored when parsing.

src/commands/edge_app.rs
```diff
@@ -0,0 +1,250 @@
...
+        map.remove("permissions");
+
+        let yaml = serde_yaml::to_string(map)?;
+        let input = File::create(path)?;
```
korvyashka, line 56: It should fail if manifest already exists - correct?

src/commands/edge_app.rs
```diff
@@ -0,0 +1,250 @@
...
+        Ok(())
+    }
+
+    pub fn publish(self, path: &Path) -> Result<EdgeAppManifest, CommandError> {
+        let url = format!("{}/v4/edge_apps", &self.authentication.config.url);
+
+        let data = fs::read_to_string(path)?;
+        // by converting to struct we make sure there no extra fields and all required fields are present
+        let edge_app: EdgeAppManifest = serde_yaml::from_str(&data)?;
+        let mut binding = serde_json::to_value(edge_app)?;
```
korvyashka, line 68: Not sure about conversion to json - you already have a struct edge_app - do you use it to be able to delete fields easily?

src/commands/edge_app.rs
```diff
@@ -0,0 +1,250 @@
...
+                response.status().as_u16(),
+            ));
+        }
+
+        // I think there is no need to check for size - this vector should always be size 1.
+        // We can let it crash in run time otherwise.
+        let updated_manifests: Vec<EdgeAppManifest> = serde_json::from_str(&response.text()?)?;
+
+        // overwrite manifest file now that we have id and extra fields.
+        let yaml = serde_yaml::to_string(&updated_manifests[0])?;
```
korvyashka, line 123: Should we also drop created_by, etc fields from manifest?

manifest data is:
```
--  Manifest data
    name         text        NOT NULL,
    version      text        NOT NULL,
    description  text        NOT NULL,
    icon         text        NOT NULL,
    author       text        NOT NULL,
    homepage_url text        NOT NULL,
    permissions  text[]      DEFAULT '{}'
```

and id.

src/commands/edge_app.rs
```diff
@@ -0,0 +1,250 @@
...
+        // init should make no requests so it's whatever for server url
+        let config = Config::new("asdf".to_owned());
+        let authentication = Authentication::new_with_config(config);
+        let command = EdgeAppCommand::new(authentication);
+
+        let p = tmp_dir.path().join("screenly.yml");
+        assert!(command.init(Path::new(p.to_str().unwrap())).is_ok());
+
+        assert_eq!(
+            EdgeAppManifest::default(),
```
korvyashka, line 159: Let's be precise here meaning raw string

```rust
#r"name:
    - some_name
description:
    - ....
"#
```

or load yaml from this string.

korvyashka (changes requested):

src/commands/edge_app.rs
sergey-borovkov: yes

src/commands/edge_app.rs
sergey-borovkov: no - they won't be ignored, parsing will faile. Though it might configurable. 
I am using them when publishing. At that point, they go to the ready file. If we don't need them I think we should remove them from the table at which point I can modify the application.

src/commands/edge_app.rs
sergey-borovkov: Hmm, yeah that was the idea. But I ended up not using that code as payload is a separate struct. I will update it.

sergey-borovkov: But parsing to struct first is still valuable as it validates that all fields are present.

src/commands/edge_app.rs
sergey-borovkov: Maybe - is that data even relevant for users all the time? can be somewhere in another table.

src/commands/edge_app.rs
sergey-borovkov: what's the purpose of that? but fine.

src/commands/edge_app.rs
korvyashka: We do need them at the db and I guess it could be usable to API user as information - but is not manifest data. I believe it could be configurable though.

src/commands/edge_app.rs
korvyashka: It is a property of edge app - no need to separate it, even in the view.
What you could do is select only required fields when doing API request:

?select=name,version,description...

https://postgrest.org/en/stable/api.html#embedding-after-insertions-updates-deletions

src/commands/edge_app.rs
korvyashka: Usually test with precise strings is more readable and is determined. 
Imagine you make a mistake in the result of default() function and response would include your mistake - in that way test will pass, but there will be an error.

But comparing it always to exact response would catch that error.

korvyashka (approved):

PR #65, opened 2023-08-11
Adds: update upload function to work with edge apps iteration 2

korvyashka:
Refs https://phorge.wireload.net/T7243

Update function now uses iteration 2 of edge apps, meaning changes:
 - Installation instead of root id
 - Settings created per app, not per revision
 - Only new settings are created, old remote are not deleted

For upload there is missing a PATCH update if setting parameters has changed.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -277,18 +277,13 @@ impl EdgeAppCommand {
         let changed_files = detect_changed_files(&local_files, &remote_files)?;
         debug!("Changed files: {:?}", &changed_files);
 
-        let remote_settings = if manifest.root_asset_id.is_empty() {
-            Vec::new()
-        } else {
-            serde_json::from_value::<Vec<Setting>>(commands::get(
+        let remote_settings = serde_json::from_value::<Vec<Setting>>(commands::get(
             &self.authentication,
             &format!(
-                "v4/edge-apps/settings?select=type,default_value,optional,title,help_text&app_id=eq.{}&app_revision=eq.{}&order=title.asc",
-                manifest.app_id,
-                manifest.revision
+                "v4/edge-apps/settings?select=type,default_value,optional,title,help_text&app_id=eq.{}&order=title.asc",
+                manifest.app_id
             ),
-        )?)?
-        };
+        )?)?;
         let changed_settings = detect_changed_settings(&manifest, &remote_settings)?;
         let file_tree = generate_file_tree(&local_files, edge_app_dir);
         let old_file_tree = self.get_file_tree(&manifest);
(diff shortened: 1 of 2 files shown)


src/commands/edge_app.rs
```diff
@@ -483,26 +470,26 @@ impl EdgeAppCommand {
         Ok(())
     }
 
-    fn get_root_edge_app_asset(&self, manifest: &EdgeAppManifest) -> Result<String, CommandError> {
+    fn check_installation_exists(&self, manifest: &EdgeAppManifest) -> Result<(), CommandError> {
```
sergey-borovkov, line 473: well it should be get_installation then because we are not going to store the installation id in the file

src/commands/edge_app.rs
```diff
@@ -635,15 +621,10 @@ impl EdgeAppCommand {
     fn install_edge_app(&self, manifest: &EdgeAppManifest) -> Result<String, CommandError> {
```
sergey-borovkov, line 621: pls run clippy

src/commands/edge_app_utils.rs
```diff
@@ -308,9 +298,43 @@ mod tests {
         assert!(result.is_ok());
         let changes = result.unwrap();
         assert!(changes.changes_detected);
-        assert_eq!(changes.creates.len(), 2);
+        assert_eq!(changes.creates.len(), 1);
+        assert_eq!(changes.creates[0].title, "google_maps_api_key");
     }
 
+    // TODO: Update test, when patching is implemented
```
sergey-borovkov, line 305: can't you disable it instead? I really hate commented out code

sergey-borovkov (changes requested):

src/commands/edge_app.rs
korvyashka: It doesn't actually return installation, as we discussed that we don't want to store it.
So what this function actually does is it checks if installation exists or return error. check or may be ensure fits it.

src/commands/edge_app_utils.rs
korvyashka: ok, though I wanted to implement it in next PR

src/commands/edge_app.rs
korvyashka: Clippy didn't find anything

src/commands/edge_app_utils.rs
korvyashka: Disabled

src/commands/edge_app.rs
korvyashka: In this PR: https://github.com/Screenly/cli/pull/67
renamed it to get_installation as I need it there.

sergey-borovkov (approved):

PR #68, opened 2023-08-16
Parallel upload

sergey-borovkov:
## What does this PR do?
Upload now does parallel upload and progress is displayed as number of files uploaded
## Checklist before merging

- [ ] If have added tests.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -16,8 +16,10 @@ use serde_yaml;
 use std::fs;
 use std::fs::File;
 use std::io::Write;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex};
 
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


src/commands/edge_app.rs
```diff
@@ -614,27 +616,39 @@ impl EdgeAppCommand {
...
+        &self,
+        manifest: &EdgeAppManifest,
+        paths: &[PathBuf],
+    ) -> Result<(), CommandError> {
+        let pb = ProgressBar::new(paths.len() as u64);
+        pb.set_message("Files uploaded:");
+        let shared_pb = Arc::new(Mutex::new(pb));
+
+        paths.par_iter().try_for_each(|path| {
+            let result = self.upload_single_asset(manifest, path, &shared_pb);
```
korvyashka, line 629: When it is 1 or 2 bigger files - progress bar will not be as granular as it was in sequential upload version. I guess we are ok with that though...

korvyashka (approved):

src/commands/edge_app.rs
sergey-borovkov: yep

PR #71, opened 2023-08-18
Adds: promote version to channel command

korvyashka:
Refs https://phorge.wireload.net/T7132

Adds versions promote command.
`screenly edge-app version promote 2  .`

Note: I now we will replace path with app_id, but this is not the point ot this pr, so added path for now.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -306,6 +306,7 @@ pub enum EdgeAppVersionCommands {
     },
     Promote {
         /// Edge app revision to promote.
+        #[arg(short, long)]
         revision: u32,
         /// Channel to promote to. Stable by default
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -312,6 +304,15 @@ pub enum EdgeAppVersionCommands {
         #[arg(short, long, action = clap::ArgAction::SetTrue)]
         json: Option<bool>,
     },
+    Promote {
+        /// Edge app revision to promote.
+        revision: u32,
```
sergey-borovkov, line 309: revision shopuld be named argument, so please add short, long.
There will be no positional arguments

sergey-borovkov (approved):

PR #75, opened 2023-08-22
Feature/wait and progress processing assets

korvyashka:
Refs https://phorge.wireload.net/T7309

Asset processing could take a while due to different reasons, and 20 secs could be a few time.
So decided to increase it significantly and add a processing progress bar.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -450,7 +450,7 @@ impl EdgeAppCommand {
                         progress_bar.set_message("Processing Items:");
                     }
                     None => {
-                        pb = Option::Some(ProgressBar::new(array.len() as u64));
+                        pb = Some(ProgressBar::new(array.len() as u64));
                         assets_to_process = array.len() as u64;
                     }
                 }
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -415,8 +415,11 @@ impl EdgeAppCommand {
         manifest: &EdgeAppManifest,
     ) -> Result<(), CommandError> {
         const SLEEP_TIME: u64 = 2;
-        const MAX_WAIT_TIME: u64 = 20; // 20 seconds
+        const MAX_WAIT_TIME: u64 = 1000; // 1000 seconds - it could take a while for assets to process
         let mut total_duration = 0;
+
+        let mut pb: Option<ProgressBar> = Option::None;
+        let mut assets_amount = 0;
 
         loop {
             // TODO: we are not handling possible errors in asset processing here.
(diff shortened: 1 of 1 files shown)


src/commands/edge_app.rs
```diff
@@ -415,9 +415,12 @@ impl EdgeAppCommand {
         manifest: &EdgeAppManifest,
     ) -> Result<(), CommandError> {
         const SLEEP_TIME: u64 = 2;
-        const MAX_WAIT_TIME: u64 = 20; // 20 seconds
+        const MAX_WAIT_TIME: u64 = 1000; // 1000 seconds - it could take a while for assets to process
         let mut total_duration = 0;
 
+        let mut pb: Option<ProgressBar> = Option::None;
+        let mut assets_amount = 0;
```
sergey-borovkov, line 422: assets_number? assets_len? I am not even sure what this var immediately means? Processed amount?

src/commands/edge_app.rs
```diff
@@ -415,9 +415,12 @@ impl EdgeAppCommand {
         manifest: &EdgeAppManifest,
     ) -> Result<(), CommandError> {
         const SLEEP_TIME: u64 = 2;
-        const MAX_WAIT_TIME: u64 = 20; // 20 seconds
+        const MAX_WAIT_TIME: u64 = 1000; // 1000 seconds - it could take a while for assets to process
```
sergey-borovkov, line 418: Maybe something like this would be better to explicitly check timeouts 
```
loop {
        if start_time.elapsed().as_secs() > MAX_WAIT_TIME {
            return Err(CommandError::AssetProcessingTimeout);
        }```

File: src/commands/edge_app.rs [hunk 3]
```diff
@@ -428,16 +431,29 @@ impl EdgeAppCommand {
...
-                    "v4/assets?select=status&app_id=eq.{}&app_revision=eq.{}&status=neq.finished&limit=1",
+                    "v4/assets?select=status&app_id=eq.{}&app_revision=eq.{}&status=neq.finished",
                     manifest.app_id, manifest.revision
                 ),
             )?;
             debug!("ensure_assets_processing_finished: {:?}", &value);
 
             if let Some(array) = value.as_array() {
                 if array.is_empty() {
+                    if let Some(_pb) = pb {
```
sergey-borovkov, line 442: if let Some(pb) = pb.as_ref() {
                    pb.finish_with_message("Assets processed");
                }
                break;

src/commands/edge_app.rs
```diff
@@ -428,16 +431,29 @@ impl EdgeAppCommand {
...
+                        _pb.finish_with_message("Assets processed");
+                    }
                     break;
                 }
+                match pb {
```
sergey-borovkov, line 447: match &mut pb {

src/commands/edge_app.rs
```diff
@@ -415,32 +415,47 @@ impl EdgeAppCommand {
...
+                    }
                     break;
                 }
+                match &mut pb {
+                    Some(ref mut progress_bar) => {
+                        progress_bar.set_length(assets_to_process - (array.len() as u64));
+                        progress_bar.set_message("Processing Items:");
+                    }
+                    None => {
+                        pb = Option::Some(ProgressBar::new(array.len() as u64));
```
sergey-borovkov, line 453: Why Option::Some? Just use Some/None - that's how i see it it used everywhere

sergey-borovkov (approved):

Issue #61, opened 2023-08-10
Resolve YAML syntax


vpetersson:
Right now, our YAML doesn't pass regular linting. In the long run we want to move to [StrictYAML](https://hitchdev.com/strictyaml/).
```
$ yamllint screenly.yml
1:1       warning  missing document start "---"  (document-start)
```

PR #78, opened 2023-08-24
Add --- at the beggining of *.y*ml files

StanIwan:
## What does this PR do?
https://phorge.wireload.net/T7291

Initially, I thought the task to be about implementing StrictYAML parser in the project (based on this issue https://github.com/Screenly/cli/issues/61), @sergey-borovkov clarified that it's just a simple fix of the issues that `yamllint` shows.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,4 +1,5 @@
 # action.yml
+---
 name: 'Screenly'
 description: 'GitHub Actions for Screenly.'
 author: 'Screenly, Inc'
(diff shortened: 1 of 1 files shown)


StanIwan:
@sergey-borovkov, should I also update files in `.github` repository?

image

StanIwan:
@sergey-borovkov running `yamllint` on the `cli` projects reports truthy warning, but it seems that it's not right, should I add a `.yamllint` config file to the project that would suppress it?

image

sergey-borovkov:
I am very confused here. GH Actions don't really matter...
That's the function that needs to be fixed. 
https://github.com/Screenly/cli/blob/a29bdacfbdf10be82d942f3bff9609f186885a29/src/commands/mod.rs#L267

action.yml
```diff
@@ -1,4 +1,5 @@
 # action.yml
+---
```
sergey-borovkov, line 2: no need.

PR #80, opened 2023-08-24
Add -- at the start of created .yml file

StanIwan:

Quick fix for https://github.com/Screenly/cli/issues/61 and https://phorge.wireload.net/T7291. Tests for veryfying the `save_to_file` function and if the required fields are present will be done under https://phorge.wireload.net/T7350.

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -1,0 +1,5 @@
+---
+extends: default
+
+rules:
+  line-length: disable
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -2,4 +2,4 @@
 extends: default
 
 rules:
-  line-length: disable
+  line-length: {max: 140}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,3 +39,5 @@ rayon = "1.7.0"
 [dev-dependencies]
 httpmock = "0.6"
 envtestkit = "1.1.2"
+tempfile = "3.8"
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ serde_yaml = "0.9.17"
 sha1 = "0.10.5"
 sha2 = "0.10.7"
 simple_logger = { version = "4.0.0", features= ["colors"]}
-tempdir = "0.3.7"
 term = "0.7.0"
 thiserror = "1.0.37"
 walkdir = "2.3.3"
@@ -40,4 +39,3 @@ rayon = "1.7.0"
 httpmock = "0.6"
 envtestkit = "1.1.2"
 tempfile = "3.8"
-
--- a/action.yml
+++ b/action.yml
@@ -1,3 +1,4 @@
+---
 # action.yml
 name: 'Screenly'
 description: 'GitHub Actions for Screenly.'
@@ -21,7 +22,9 @@ runs:
     - id: download-cli
       run: |
         wget -q -O /tmp/screenly-cli.tar.gz \
-          "https://github.com/Screenly/cli/releases/download/${{ inputs.cli_version }}/screenly-cli-x86_64-unknown-linux-gnu.tar.gz"
+          "https://github.com/Screenly/cli/releases/download/\
+          ${{ inputs.cli_version }}/\
+          screenly-cli-x86_64-unknown-linux-gnu.tar.gz"
         cd /tmp
         tar xfz screenly-cli.tar.gz
         chmod +x screenly
--- a/action.yml
+++ b/action.yml
@@ -22,9 +22,7 @@ runs:
     - id: download-cli
       run: |
         wget -q -O /tmp/screenly-cli.tar.gz \
-          "https://github.com/Screenly/cli/releases/download/\
-          ${{ inputs.cli_version }}/\
-          screenly-cli-x86_64-unknown-linux-gnu.tar.gz"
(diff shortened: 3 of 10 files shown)


action.yml
```diff
@@ -21,7 +22,9 @@ runs:
     - id: download-cli
       run: |
         wget -q -O /tmp/screenly-cli.tar.gz \
-          "https://github.com/Screenly/cli/releases/download/${{ inputs.cli_version }}/screenly-cli-x86_64-unknown-linux-gnu.tar.gz"
+          "https://github.com/Screenly/cli/releases/download/\
```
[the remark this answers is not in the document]
sergey-borovkov: well let's increase line limit then. We are increasing it everywhere else.

src/commands/mod.rs
```diff
@@ -267,7 +267,7 @@ impl EdgeAppManifest {
     pub fn save_to_file(manifest: &EdgeAppManifest, path: &Path) -> Result<(), CommandError> {
         let yaml = serde_yaml::to_string(&manifest)?;
         let manifest_file = File::create(path)?;
-        write!(&manifest_file, "{yaml}")?;
+        write!(&manifest_file, "---\n{}", yaml)?;
```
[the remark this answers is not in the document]
sergey-borovkov: they still should be ideally in this PR - so that we have a test for modified behavior. Those tests in that task will not be related. And that "linting" I don't think is in any way related to yaml linter. That linter will validate if the values used are correct.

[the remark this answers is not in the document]
sergey-borovkov: SO my point is that the second task is completely unrelated - save_to_file - there should not be modified at all as well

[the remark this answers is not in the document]
StanIwan: > they still should be ideally in this PR - so that we have a test for modified behavior. Those tests in that task will not be related. And that "linting" I don't think is in any way related to yaml linter. That linter will validate if the values used are correct.

They will be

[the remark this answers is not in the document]
StanIwan: Ok, so it's much simpler then.

I'll add tests and re-request the review.

src/commands/mod.rs
```diff
@@ -586,3 +586,41 @@ impl Formatter for PlaylistItems {
...
+            ..Default::default()
+        };
+
+        EdgeAppManifest::save_to_file(&manifest, &file_path).unwrap();
+
+        let mut file = File::open(&file_path).unwrap();
+        let mut contents = String::new();
+        file.read_to_string(&mut contents).unwrap();
+
+        let expected_yaml = serde_yaml::to_string(&manifest).unwrap();
```
sergey-borovkov, line 619: please write the whole content by hand (or copy paster it from that file here)

StanIwan: sure

Cargo.toml
```diff
@@ -39,3 +39,5 @@ rayon = "1.7.0"
 [dev-dependencies]
 httpmock = "0.6"
 envtestkit = "1.1.2"
+tempfile = "3.8"
```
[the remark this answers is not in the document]
StanIwan: It is possible to move it to dev-deps. In such case I would suggest switching to `tempfile` cargo as it seems to be kept alive and "supported" (I guess)
image

`tempdir` cargo that is used now was updated 5 years ago:
image

It's a simple switch.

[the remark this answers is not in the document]
sergey-borovkov: Ok sure - do the switch

.yamllint.yml
```diff
@@ -0,0 +1,5 @@
+---
+extends: default
+
+rules:
+  line-length: disable
```
sergey-borovkov, line 5: don't disable it - set value higher than we had. Try 120/160

src/commands/mod.rs
```diff
@@ -586,3 +586,54 @@ impl Formatter for PlaylistItems {
...
+    type: string
+    default_value: stranger
+    title: username
+    optional: true
+    help_text: An example of a setting that is used in index.html
+"#;
+
+        assert_eq!(contents, expected_contents);
+
+        dir.close().unwrap();
```
sergey-borovkov, line 637: why do you need to call close manually?

src/signature.rs
```diff
@@ -60,8 +60,8 @@ pub fn checksum(chunk: &[u8]) -> Vec<u8> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::fs;
-    use tempdir::TempDir;
+    use std::fs;
```
sergey-borovkov, line 63: please run cargo fmt before pushing

StanIwan: I'll remember that for the future

src/commands/mod.rs
```diff
@@ -586,3 +586,54 @@ impl Formatter for PlaylistItems {
...
+                help_text: "An example of a setting that is used in index.html".to_string(),
+            }],
+            ..Default::default()
+        };
+
+        EdgeAppManifest::save_to_file(&manifest, &file_path).unwrap();
+
+        let mut file = File::open(&file_path).unwrap();
+        let mut contents = String::new();
+        file.read_to_string(&mut contents).unwrap();
```
sergey-borovkov, line 617: you can read to string in one liner.

let contents = fs::read_to_string(file_path)

src/commands/mod.rs
StanIwan: I think that it's a good practice and makes tests more robust. 

I've seen it it at leas one other test - should I delete it?

sergey-borovkov (approved):

PR #84, opened 2023-08-25
Capture stdout and export it

vpetersson:
This is needed to use the output in other commands, such as parsing the releases.


StanIwan (approved):
LGTM

sergey-borovkov (approved):

PR #86, opened 2023-08-27
T7350 yml verification script

StanIwan:
## What does this PR do?

This PR adds a `screenly.yml` verification script that uses schema similar to one that `snapcraft` uses (https://github.com/snapcore/snapcraft/blob/main/schema/snapcraft.json).

Script can be used to check if all required fields are present in `screenly.yml` and if their type is correct. The script returns exit code according to verification result for use in the CI and also prints all issues found in the `.yml. file.

## How has this been tested?
- Manual tests with `screenly.yml` that is created by the `CLI`

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,3 +139,52 @@ with:
   screenly_api_token: ${{ secrets.SCREENLY_API_TOKEN }}
   cli_commands: screen list
 ```
+
+## Verifying YAML Configuration
+
+### Introduction
+The `verify_screenly_yml.sh` script is a utility that validates the `screenly.yml` file against a predefined JSON schema. This ensures that your configuration adheres to the expected format before deploying or using the application.
+
+### Prerequisites
+Before running the script, make sure you have the following installed:
+
+- `Python 3.x`
+- `jsonschema` Python package
+
+You can install the jsonschema package using pip:
+
+```bash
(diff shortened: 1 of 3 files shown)


StanIwan:
@renatgalimov - I've used the approach that you advised, my main concerns here are:
- verification script requires `python` - I don't think that this can be easily done using bash alone
- verification script requires `jsonschema` Python package - which must be manually installed
- script assumes that it will be run from the `cli` directory - this could be fixed by 2 arguments (path to .yml file; path to schema) instead of current 1 (path to .yml file)

Please let me know what you think, if my changes are sufficient, I'll make this a `Ready for review` PR

renatgalimov:
@StanIwan , I like it!

StanIwan:
> Why is this in Python? The CLI is written in Rust. We want to use it like this `screenly edge-app lint`. Not a separate tool.

`lint in rust` and `lint as a command in the cli` was the info I was missing, I'll implement it as a rust function that can be also called as a cli command.

sergey-borovkov (changes requested):
I guess this PR should be thrown away.

StanIwan:
> I guess this PR should be thrown away.

I'll just replace bash script with python with rust, serde crate allows me to use my current json schema and verify yaml.

PR #87, opened 2023-08-28
Adds: check undefined secrets on promote.

korvyashka:
Refs https://phorge.wireload.net/T7376

Before promote we do check if some secrets are undefined.
```
screenly edge-app version promote --revision 1
Failed to promote edge app version: Warning: these secrets are undefined: ["greeting1","greeting2"]..
```

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -304,7 +304,14 @@ impl EdgeAppCommand {
         revision: &u32,
         channel: &String,
     ) -> Result<(), CommandError> {
-        self.ensure_secrets_defined(app_id)?;
+        let secrets = self.get_undefined_secrets(app_id)?;
+        if !secrets.is_empty() {
+            return Err(CommandError::WarningUndefinedSecrets(
+                serde_json::to_string(&secrets)?,
+            ));
+        }
+
+        debug!("No undefined secrets found");
 
         let response = commands::patch(
             &self.authentication,
@@ -335,7 +342,7 @@ impl EdgeAppCommand {
         Ok(())
     }
 
-    fn ensure_secrets_defined(&self, app_id: &str) -> Result<(), CommandError> {
+    fn get_undefined_secrets(&self, app_id: &str) -> Result<Vec<String>, CommandError> {
         let installation_id = self.get_or_create_installation(app_id)?;
 
         let undefined_secrets_response = commands::get(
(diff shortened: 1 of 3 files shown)


src/commands/edge_app.rs
```diff
@@ -333,6 +335,29 @@ impl EdgeAppCommand {
...
+
+        let undefined_secrets_response = commands::get(
+            &self.authentication,
+            &format!(
+                "v4/edge-apps/secrets/undefined?installation_id={}",
+                installation_id
+            ),
+        )?;
+
+        let titles = serde_json::from_value::<Vec<String>>(undefined_secrets_response)?;
```
sergey-borovkov, line 349: to be honest maybe returning the list of undefined secrets would be better? Looks more natural. You return it in error, but we return a lot of different types of errors.
Returning the list of undefined secrets and checking that its length is 0 sounds better to me. What do you think?

src/commands/edge_app.rs
```diff
@@ -1583,8 +1658,89 @@ mod tests {
...
+        };
+
+        let result = command.promote_version(&manifest.app_id, &7, &"public".to_string());
+
+        installation_mock.assert();
+        installation_mock_create.assert();
+        undefined_secrets_mock.assert();
+
+        assert!(!&result.is_ok());
+        assert!(result.unwrap_err().to_string().contains("Warning: these secrets are undefined: [\"undefined_secret\",\"another_undefined_secret\"]. Use --ignore-warning to ignore"));
```
sergey-borovkov: this looks outdated :) That's one of the reasons I don't like it as much as well. Modifying stuff won't immediately break because it's inside of the string

korvyashka: I already updated it)

src/commands/edge_app.rs
korvyashka: @sergey-borovkov what you mean is that it should not be an error, but just a list response?

src/commands/edge_app.rs
```diff
@@ -304,6 +304,15 @@ impl EdgeAppCommand {
...
         channel: &String,
     ) -> Result<(), CommandError> {
+        let secrets = self.get_undefined_secrets(app_id)?;
+        if !secrets.is_empty() {
+            return Err(CommandError::WarningUndefinedSecrets(
+                serde_json::to_string(&secrets)?,
+            ));
+        }
+
+        debug!("No undefined secrets found");
```
sergey-borovkov, line 314: "All secrets are defined." better maybe?

src/commands/edge_app.rs
```diff
@@ -304,6 +304,15 @@ impl EdgeAppCommand {
         revision: &u32,
         channel: &String,
     ) -> Result<(), CommandError> {
+        let secrets = self.get_undefined_secrets(app_id)?;
+        if !secrets.is_empty() {
+            return Err(CommandError::WarningUndefinedSecrets(
```
sergey-borovkov, line 309: It's not a warning now though :)

sergey-borovkov (approved):

PR #88, opened 2023-08-28
Validate yaml manifest

StanIwan:
## What does this PR do?
Attempts to "Lint" screenly.yml
Based on changes from https://github.com/Screenly/Screenly/pull/350 following fields are not longer required:
- user_version
- description
- icon
- author
- homepage_url

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -101,6 +101,8 @@ pub enum CommandError {
     FileSystemError(String),
     #[error("Asset processing timeout")]
     AssetProcessingTimeout,
+    #[error("Error validating manifest file")]
+    ValidationError(String),
 }
 
 pub fn get(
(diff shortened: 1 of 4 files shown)


StanIwan:
@sergey-borovkov - please tell me if it goes in the right direction :)

src/commands/mod.rs
```diff
@@ -271,6 +272,49 @@ impl EdgeAppManifest {
...
+                let mut errors = Vec::new();
+    
+                if manifest.app_id.is_empty() {
+                    errors.push("app_id is empty".to_string());
+                }
+    
+                if manifest.settings.is_empty() {
+                    errors.push("settings are empty".to_string());
+                } else {
+                    for (i, setting) in manifest.settings.iter().enumerate() {
```
sergey-borovkov, line 294: This is not a good approach. I don't like it. Much more natural is to make a custom deserializer that will not accept an empty string.

sergey-borovkov (changes requested):

PR #89, opened 2023-08-29
Added: a new create option for create edge-app with existing directory

rusko124:
## What does this PR do?
Added a new option for create command: `screenly edge-app create --in-place`.

This option allows creating edge-app using an existing edge-app directory (see Playground edge-apps for an example).

@vpetersson please add a ticket number. I can't find it.

## How has this been tested?
Manually and unit tests.

## Checklist before merging

- [x] If have added tests.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -112,7 +112,7 @@ impl EdgeAppCommand {
         let mut manifest: EdgeAppManifest = serde_yaml::from_str(&data)?;
 
         if !manifest.app_id.is_empty() {
-            return Err(CommandError::FileSystemError("app_id in screenly.yml must be empty".to_string()));
+            return Err(CommandError::FileSystemError("app_id in screenly.yml should not be set".to_string()));
         }
 
         let response = commands::post(
@@ -984,7 +984,7 @@ mod tests {
         assert!(result.is_err());
         assert_eq!(
             result.unwrap_err().to_string(),
-            "Filesystem error: app_id in screenly.yml must be empty"
+            "Filesystem error: app_id in screenly.yml should not be set"
         );
     }
 
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -36,6 +36,14 @@ pub struct AssetSignature {
     pub(crate) signature: String,
 }
 
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct EdgeAppCreationResponse {
+    #[serde(default)]
+    pub id: String,
+    #[serde(default)]
+    pub name: String,
+}
+
 impl EdgeAppCommand {
     pub fn new(authentication: Authentication) -> Self {
         Self { authentication }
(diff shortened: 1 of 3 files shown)


src/commands/edge_app.rs
```diff
@@ -95,6 +95,54 @@ impl EdgeAppCommand {
...
+            return Err(CommandError::FileSystemError(format!(
+                "The directory {} should contain screenly.yml and index.html files",
+                parent_dir_path.display()
+            )));
+        }
+
+        let data = fs::read_to_string(path)?;
+        let mut manifest: EdgeAppManifest = serde_yaml::from_str(&data)?;
+
+        if manifest.app_id != "" {
```
sergey-borovkov, line 114: Nope app_id can't be empty. We are going to not allow that. Either it's not present there or non empty. Though I guess at the moment it works

sergey-borovkov: You can change it now to be an Option<String> in manifest though leaving non empty part. If option is None your code should work

src/commands/edge_app.rs
```diff
@@ -95,6 +95,52 @@ impl EdgeAppCommand {
         Ok(())
     }
 
+    pub fn create_in_place(&self, name: &str, path: &Path) -> Result<(), CommandError> {
```
rusko124, line 98: I don't want to do this in create because there will literally be if-else everywhere

src/commands/edge_app.rs
```diff
@@ -95,6 +95,52 @@ impl EdgeAppCommand {
...
+                "The directory {} should contain screenly.yml and index.html files",
+                parent_dir_path.display()
+            )));
+        }
+
+        let data = fs::read_to_string(path)?;
+        let mut manifest: EdgeAppManifest = serde_yaml::from_str(&data)?;
+
+        if !manifest.app_id.is_empty() {
+            return Err(CommandError::FileSystemError("app_id in screenly.yml should not be set".to_string()));
```
sergey-borovkov, line 115: Error is confusing though. Because if will try to remove it - still won't work. Also it's not a file system error

src/commands/edge_app.rs
```diff
@@ -95,6 +95,52 @@ impl EdgeAppCommand {
...
+        }
+
+        let response = commands::post(
+            &self.authentication,
+            "v4/edge-apps?select=id,name",
+            &json!({ "name": name }),
+        )?;
+
+        #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
```
sergey-borovkov, line 124: Please move this struct out of function since it's used in another function as well. The code below looks like a duplicate of what create does as well

src/commands/edge_app.rs
```diff
@@ -827,7 +873,119 @@ mod tests {
...
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("should contain screenly.yml and index.html files")
+        );
+    }
+
+    #[test]
+    fn test_init_edge_app_when_manifest_has_non_empty_app_id_should_return_error() {
```
sergey-borovkov, line 959: test_create_in_place?

src/commands/edge_app.rs
sergey-borovkov: that's fine but there is a lot of common code, no?

rusko124: I think these functions have 50% of the same code. But anyhow It will be ugly.

PR #90, opened 2023-08-29
Adds: check for processing error of assets.

korvyashka:
Refs https://phorge.wireload.net/T7389

When there was an error with processing assets, cli will be timeout after 1000 seconds, while we actually now that there was an error during processing.
Fail immediately on error and print it.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -453,13 +453,24 @@ impl EdgeAppCommand {
             let value = commands::get(
                 &self.authentication,
                 &format!(
-                    "v4/assets?select=status&app_id=eq.{}&app_revision=eq.{}&status=neq.finished",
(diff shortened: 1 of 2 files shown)


src/commands/edge_app.rs
```diff
@@ -453,13 +453,24 @@ impl EdgeAppCommand {
...
                     app_id, revision
                 ),
             )?;
             debug!("ensure_assets_processing_finished: {:?}", &value);
 
             if let Some(array) = value.as_array() {
+                for item in array {
+                    if let Some(status) = item["status"].as_str() {
+                        if status == "error" {
+                            return Err(CommandError::AssetProcessingError(format!(
```
sergey-borovkov, line 466: Let's maybe switch to anyhow for new code ? We already have a gazillion of error types and it's not helpful anywhere to be honest

sergey-borovkov (approved):

PR #92, opened 2023-08-30
Fix: filter out edge-app-files from assets list command

rusko124:
## What does this PR do?
filter out edge-app-files from assets list command. Just the same way as we do it in web-console.

https://phorge.wireload.net/T7310

## How has this been tested?
Manually

## Checklist before merging

- [ ] If have added tests.

### DIFF
--- a/src/commands/asset.rs
+++ b/src/commands/asset.rs
@@ -23,7 +23,7 @@ impl AssetCommand {
     pub fn list(&self) -> anyhow::Result<Assets, CommandError> {
         Ok(Assets::new(commands::get(
             &self.authentication,
-            "v4/assets",
+            "v4/assets?type=in.('appweb','audio','edge-app','image','video','web')",
         )?))
     }
 
@@ -214,6 +214,7 @@ mod tests {
         mock_server.mock(|when, then| {
             when.method(GET)
                 .path("/v4/assets")
+                .query_param("type", "in.('appweb','audio','edge-app','image','video','web')")
                 .header("Authorization", "Token token")
                 .header(
                     "user-agent",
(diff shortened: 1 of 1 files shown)


src/commands/asset.rs
```diff
@@ -23,7 +23,7 @@ impl AssetCommand {
     pub fn list(&self) -> anyhow::Result<Assets, CommandError> {
         Ok(Assets::new(commands::get(
             &self.authentication,
-            "v4/assets",
+            "v4/assets?type=in.('appweb','audio','edge-app','image','video','web')",
```
korvyashka, line 26: lets remove appweb - they are not used

korvyashka (approved):

src/commands/asset.rs
sergey-borovkov, line 26: Why not say everything but edge app files ? If we add new asset type it won't be shown by default

rusko124: > If we add new asset type it won't be shown by default

Yes, we want exactly this approach.

src/commands/asset.rs
```diff
@@ -214,6 +214,7 @@ mod tests {
         mock_server.mock(|when, then| {
             when.method(GET)
                 .path("/v4/assets")
+                .query_param("type", "in.('appweb','audio','edge-app','image','video','web')")
```
rusko124, line 217:
```suggestion
                .query_param("type", "in.('audio','edge-app','image','video','web')")
```

src/commands/asset.rs
rusko124, line 26:
```suggestion
            "v4/assets?type=in.('audio','edge-app','image','video','web')",
```

korvyashka: That should be deleted from appserver too - let's not increase tech debt.

vpetersson: We should remove `appweb`. But really, we should filter out edge-app files in the API rather than as a filter.

PR #94, opened 2023-08-31
Adds: delete edge-app command - same as screen deletion

korvyashka:
Refs https://phorge.wireload.net/T7415

Same as screen deletion - ask for name promt before deletion

`screenly edge-app delete`

Also updated manifest to have app_id optional -> `Option<String>`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -275,7 +275,7 @@ pub enum AssetCommands {
 pub enum EdgeAppCommands {
     /// Creates Edge App in the store.
     Create {
-        /// Edge app name
+        /// Edge App name
         #[arg(short, long)]
         name: String,
         /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
@@ -286,7 +286,7 @@ pub enum EdgeAppCommands {
         in_place: Option<bool>,
     },
 
-    /// Lists your edge apps.
+    /// Lists your Edge Apps.
     List {
         /// Enables JSON output.
         #[arg(short, long, action = clap::ArgAction::SetTrue)]
@@ -336,23 +336,23 @@ pub enum EdgeAppCommands {
     #[command(subcommand)]
     Secret(EdgeAppSecretsCommands),
 
-    /// Uploads assets and settings of the edge app.
+    /// Uploads assets and settings of the Edge App.
     Upload {
         /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
         #[arg(short, long)]
         path: Option<String>,
 
-        /// Edge app id. If not specified CLI will use the id from the manifest.
+        /// Edge App id. If not specified CLI will use the id from the manifest.
         #[arg(short, long)]
         app_id: Option<String>,
     },
-    /// Deletes an edge app. This cannot be undone.
+    /// Deletes an Edge App. This cannot be undone.
     Delete {
         /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
         #[arg(short, long)]
         path: Option<String>,
 
-        /// Edge app id. If not specified CLI will use the id from the manifest.
+        /// Edge App id. If not specified CLI will use the id from the manifest.
         #[arg(short, long)]
         app_id: Option<String>,
     },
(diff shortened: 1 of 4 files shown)


vpetersson (commented):
What happens to the files on disk? Do we wipe the app id from the yaml file?

korvyashka:
@vpetersson discussed with @sergey-borovkov and decided not to do it.
Let me know if we actually want to do this.

vpetersson:
What's the rational? We currently don't enforce app name to be unique, right?

vpetersson:
Right so then we can't use this approach. We much change it to use the app id. Id say @MuratUrsavas suggestion plus app id seems reasonable.

korvyashka:
@vpetersson I'm not sure I'm following.

- suggestion from @MuratUrsavas was in another PR, https://github.com/Screenly/cli/pull/95 and it is cause PR description was wrong - update uses '--name=' of course not 'name=', so it is not relevant.

- Deletion already happens by app_id - either '--app_id=' or if not passed, from manifest in current folder. Name input is just to confirm deletion and not allow it happen instantaneusly when calling delete command for instance accidently.

- So only question left is if we need to delete screenly.yml on delete or delete app_id there or delete whole folder.
@sergey-borovkov said we shouldn't, I don't have strong opinion here - let me know if we should.

sergey-borovkov:
> Right so then we can't use this approach. We much change it to use the app id. Id say @MuratUrsavas suggestion plus app id seems reasonable.

The name is just for confirmation. You delete with id.

I don't think we should remove anything from the disk. Makes no sense to me to be honest.

src/cli.rs
```diff
@@ -299,6 +299,16 @@ pub enum EdgeAppCommands {
         #[arg(short, long)]
         path: Option<String>,
 
+        /// Edge app id. If not specified CLI will use the id from the manifest.
+        #[arg(short, long)]
+        app_id: Option<String>,
+    },
+    /// Deletes an edge app. This cannot be undone.
```
sergey-borovkov, line 306: Please capitalize. Edge App. Everywhere. If you see it incorrectly nearby please ifx it as well

src/cli.rs
```diff
@@ -299,6 +299,16 @@ pub enum EdgeAppCommands {
...
+    Delete {
+        /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
+        #[arg(short, long)]
+        path: Option<String>,
+
         /// Edge app id. If not specified CLI will use the id from the manifest.
```
sergey-borovkov, line 312: Like here

src/cli.rs
```diff
@@ -299,6 +299,16 @@ pub enum EdgeAppCommands {
         #[arg(short, long)]
         path: Option<String>,
 
+        /// Edge app id. If not specified CLI will use the id from the manifest.
+        #[arg(short, long)]
+        app_id: Option<String>,
+    },
+    /// Deletes an edge app. This cannot be undone.
+    Delete {
```
sergey-borovkov, line 307: Why is it not taking id optionally?

src/cli.rs
```diff
@@ -946,6 +956,34 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                 }
             }
         },
+        EdgeAppCommands::Delete { path, app_id } => {
+            let actual_app_id = get_actual_app_id(app_id, path);
+            match edge_app_command.get_app_name(&actual_app_id) {
+                Ok(name) => {
+                    info!("You are about to delete the app named \"{}\".  This operation cannot be reversed.", name);
```
sergey-borovkov, line 963: Edge App named maybe here as well for consistency?

sergey-borovkov (changes requested):

src/cli.rs
korvyashka: /// Edge App id. If not specified CLI will use the id from the manifest.
        #[arg(short, long)]
        app_id: Option<String>,

korvyashka: It is

sergey-borovkov: my bad, it's because how diff shows - did not notice

sergey-borovkov (approved):

PR #95, opened 2023-08-31
Adds: update name on the edge-app

korvyashka:
Refs https://phorge.wireload.net/T7416

Update name on the edge app.
Decided that cli command should look like this.

`screenly edge-app update name=new name`

Please comment on this.


MuratUrsavas:
Why not use the same format as crate command like `--name=<new_name>`?

korvyashka:
@vpetersson it uses app_id as source - either from manifest or by passing - as all other commands.

sergey-borovkov (approved):

PR #96, opened 2023-09-01
Custom deserializers for empty string and settings type

StanIwan:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,3 @@ Cargo.lock
 # Added by cargo
 
 /target
-**/screenly.yml
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ Cargo.lock
 # Added by cargo
 
 /target
+**/screenly.yml
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -514,6 +514,9 @@ fn get_actual_app_id(
     path: &Option<String>,
 ) -> Result<String, CommandError> {
     match app_id {
+        Some(id) if id.is_empty() => {
+            Err(CommandError::EmptyAppId("this".to_string()))
+        },
         Some(id) => Ok(id.clone()),
         None => {
             let manifest =
@@ -1144,7 +1147,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                     println!("Manifest file is valid.");
                 },
                 Err(e) => {
-                    println!("{}", e);
+                    println!("{e}");
                     std::process::exit(1);
                 }
             }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1139,16 +1139,12 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
         },
         EdgeAppCommands::Validate { path } => {
             let manifest_path = transform_edge_app_path_to_manifest(path);
-            match EdgeAppManifest::validate_file(&manifest_path) {
-                Ok(true) => {
+            match EdgeAppManifest::ensure_manifest_is_validated(&manifest_path) {
+                Ok(()) => {
                     println!("Manifest file is valid.");
                 },
-                Ok(false) => {
-                    println!("Manifest file is invalid.");
-                    std::process::exit(1);
-                },
                 Err(e) => {
-                    println!("Failed to validate manifest file: {}", e);
+                    println!("{}", e);
                     std::process::exit(1);
                 }
             }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -683,7 +683,7 @@ pub fn handle_cli_playlist_command(command: &PlaylistCommands) {
                     println!("{}", pretty_playlist_file);
                 }
                 Err(e) => {
-                    println!("Error occurred when getting playlist: {e:?}")
+                    eprintln!("Error occurred when getting playlist: {e:?}")
                 }
             }
         }
@@ -692,7 +692,7 @@ pub fn handle_cli_playlist_command(command: &PlaylistCommands) {
                 println!("Playlist deleted successfully.");
             }
             Err(e) => {
-                println!("Error occurred when deleting playlist: {e:?}")
+                eprintln!("Error occurred when deleting playlist: {e:?}")
             }
         },
         PlaylistCommands::Append {
@@ -738,7 +738,7 @@ pub fn handle_cli_playlist_command(command: &PlaylistCommands) {
                     println!("Playlist updated successfully.");
                 }
                 Err(e) => {
-                    println!("Error occurred when updating playlist: {e:?}")
+                    eprintln!("Error occurred when updating playlist: {e:?}")
                 }
             }
         }
@@ -913,7 +913,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                     println!("Edge app successfully created.");
                 }
                 Err(e) => {
-                    println!("Failed to publish edge app manifest: {e}.");
+                    eprintln!("Failed to publish edge app manifest: {e}.");
                     std::process::exit(1);
                 }
             }
@@ -934,7 +934,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                     );
                 }
                 Err(e) => {
-                    println!("Failed to upload edge app: {e}.");
+                    eprintln!("Failed to upload edge app: {e}.");
                     std::process::exit(1);
                 }
             }
@@ -971,7 +971,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                         println!("Edge app version successfully promoted.");
                     }
                     Err(e) => {
-                        println!("Failed to promote edge app version: {e}.");
+                        eprintln!("Failed to promote edge app version: {e}.");
                         std::process::exit(1);
                     }
                 }
@@ -1009,7 +1009,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                         println!("Edge app setting successfully set.");
                     }
                     Err(e) => {
-                        println!("Failed to set edge app setting: {}", e);
+                        eprintln!("Failed to set edge app setting: {}", e);
                         std::process::exit(1);
                     }
                 }
@@ -1047,7 +1047,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                         println!("Edge app secret successfully set.");
                     }
                     Err(e) => {
-                        println!("Failed to set edge app secret: {}", e);
+                        eprintln!("Failed to set edge app secret: {}", e);
                         std::process::exit(1);
                     }
                 }
@@ -1114,7 +1114,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                     println!("Edge app successfully updated.");
                 }
                 Err(e) => {
-                    println!("Failed to update edge app: {e}.");
+                    eprintln!("Failed to update edge app: {e}.");
                     std::process::exit(1);
                 }
             }
@@ -1130,7 +1130,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                 None => env::current_dir().unwrap(),
             };
             if !path.join(MOCK_DATA_FILENAME).exists() {
-                println!("Error: No mock-data exist. Please run \"screenly edge-app generate-mock-data\" and try again.");
+                eprintln!("Error: No mock-data exist. Please run \"screenly edge-app generate-mock-data\" and try again.");
                 std::process::exit(1);
             }
 
@@ -1147,7 +1147,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                     println!("Manifest file is valid.");
                 },
                 Err(e) => {
-                    println!("{e}");
+                    eprintln!("{e}");
                     std::process::exit(1);
                 }
             }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -356,7 +356,7 @@ pub enum EdgeAppCommands {
         #[arg(short, long)]
         app_id: Option<String>,
     },
-    /// Validates screenly.yml manifest file
+    /// Validates Edge App manifest file
     Validate {
         /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
         #[arg(short, long)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -522,8 +522,8 @@ fn get_actual_app_id(
             let manifest =
                 EdgeAppManifest::new(transform_edge_app_path_to_manifest(path).as_path()).unwrap();
             match manifest.app_id {
-                Some(id) => Ok(id),
-                None => {
+                Some(id) if !id.is_empty() => Ok(id.clone()),
+                _ =>  {
                     error!("Edge app id is not specified. Please specify it using --app-id option or add it to the manifest.");
                     Err(CommandError::MissingAppId)
                 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -515,7 +515,7 @@ fn get_actual_app_id(
 ) -> Result<String, CommandError> {
     match app_id {
         Some(id) if id.is_empty() => {
-            Err(CommandError::EmptyAppId("this".to_string()))
+            Err(CommandError::EmptyAppId)
         },
         Some(id) => Ok(id.clone()),
         None => {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -356,6 +356,12 @@ pub enum EdgeAppCommands {
         #[arg(short, long)]
         app_id: Option<String>,
     },
+    /// Validates screenly.yml manifest file
+    Validate {
+        /// Path to the directory with the manifest. If not specified CLI will use the current working directory.
+        #[arg(short, long)]
+        path: Option<String>,
+    },
 }
 
 #[derive(Subcommand, Clone, PartialEq, Eq, PartialOrd, Ord)]
@@ -1130,6 +1136,22 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
         EdgeAppCommands::GenerateMockData { path } => {
             let manifest_path = transform_edge_app_path_to_manifest(path);
             edge_app_command.generate_mock_data(&manifest_path).unwrap();
+        },
+        EdgeAppCommands::Validate { path } => {
+            let manifest_path = transform_edge_app_path_to_manifest(path);
+            match EdgeAppManifest::validate_file(&manifest_path) {
+                Ok(true) => {
+                    println!("Manifest file is valid.");
+                },
+                Ok(false) => {
+                    println!("Manifest file is invalid.");
+                    std::process::exit(1);
+                },
+                Err(e) => {
+                    println!("Failed to validate manifest file: {}", e);
+                    std::process::exit(1);
+                }
+            }
         }
     }
 }
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -325,7 +325,7 @@ impl EdgeAppCommand {
         // override app_id if user passed it
         if let Some(id) = app_id {
             if id.is_empty() {
-                return Err(CommandError::EmptyAppId("upload".to_string()));
+                return Err(CommandError::EmptyAppId);
             }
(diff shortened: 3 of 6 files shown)


src/commands/mod.rs
```diff
@@ -226,11 +231,24 @@ pub struct EdgeAppManifest {
     pub settings: Vec<Setting>,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+pub enum EdgeAppSettingType {
```
sergey-borovkov, line 235: Just call it Setting type

src/commands/mod.rs
```diff
@@ -226,11 +231,24 @@ pub struct EdgeAppManifest {
...
+        EdgeAppSettingType::String
+    }
+}
+
 // maybe we can use a better name as we have EdgeAppSettings which is the same but serde_json::Value inside
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 pub struct Setting {
     #[serde(rename = "type")]
-    pub type_: String,
+    #[serde(deserialize_with = "deserialize_setting_type")]
```
sergey-borovkov, line 250: write it in one #[serde...] block with ,

src/commands/mod.rs
```diff
@@ -268,6 +286,30 @@ where
     map.end()
 }
 
+fn deserialize_string_not_empty<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: String = Deserialize::deserialize(deserializer)?;
+    if s.is_empty() {
+        return Err(serde::de::Error::custom("String field must not be empty"))
```
sergey-borovkov, line 295: will this field name?

.gitignore
```diff
@@ -13,3 +13,4 @@ Cargo.lock
 # Added by cargo
 
 /target
+**/screenly.yml
```
sergey-borovkov, line 16: hmm?

sergey-borovkov: but there is no screenly.yml anywhere in this repo - no need I think

src/commands/mod.rs
```diff
@@ -664,11 +728,11 @@ mod tests {
 
         let expected_contents = r#"---
 app_id: test_app
-user_version: ''
-description: ''
-icon: ''
-author: ''
-homepage_url: ''
+user_version: null
+description: null
```
sergey-borovkov, line 732: What is this ? We don't support null. We only support strings... or ut should not be in the manifest file

sergey-borovkov (changes requested):
1) needs unit tests 
2) needs to handle values being none in the payload 
3) there is no validate command
4) needs to check that app id passed to functions is not empty or show an error in that case

src/commands/mod.rs
```diff
@@ -268,6 +285,53 @@ where
     map.end()
 }
 
+fn deserialize_string_not_empty<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: String = Deserialize::deserialize(deserializer)?;
+    if s.is_empty() {
+        return Err(serde::de::Error::custom("String cannot be empty"))
```
sergey-borovkov, line 294: I asked before - will this show field name ?

sergey-borovkov (changes requested):

src/commands/mod.rs
StanIwan: I've replied back then (https://github.com/Screenly/cli/pull/96#discussion_r1313016497): I don't think that you can get field name here, you get line and column.

src/commands/mod.rs
StanIwan: So if we have EdgeAppManifest like this (default for optional struct fields):

```       
 let manifest = EdgeAppManifest {
            app_id: "test_app".to_string(),
            settings: vec![Setting {
                title: "username".to_string(),
                type_: SettingType::String,
                default_value: "stranger".to_string(),
                optional: true,
                help_text: "An example of a setting that is used in index.html".to_string(),
            }],
            ..Default::default()
        };
```
It should not write this fields to the `screenly.yml`?

sergey-borovkov: > So if we have EdgeAppManifest like this (default for optional struct fields):
> 
> ```
>  let manifest = EdgeAppManifest {
>             app_id: "test_app".to_string(),
>             settings: vec![Setting {
>                 title: "username".to_string(),
>                 type_: SettingType::String,
>                 default_value: "stranger".to_string(),
>                 optional: true,
>                 help_text: "An example of a setting that is used in index.html".to_string(),
>             }],
>             ..Default::default()
>         };
> ```
> 
> It should not write this fields to the `screenly.yml`?

If they are empty it should not.

sergey-borovkov: Update after daily: not going to check that those Optional fields are not empty (except for app_id -> app_id should still be checked for length)

src/commands/edge_app.rs
```diff
@@ -1242,11 +1242,11 @@ mod tests {
         let command = EdgeAppCommand::new(authentication);
         let manifest = EdgeAppManifest {
             app_id: "01H2QZ6Z8WXWNDC0KQ198XCZEW".to_string(),
-            user_version: "1".to_string(),
-            description: "asdf".to_string(),
-            icon: "asdf".to_string(),
-            author: "asdf".to_string(),
-            homepage_url: "asdfasdf".to_string(),
+            user_version: Some("1".to_string()),
```
sergey-borovkov, line 1245: This is not related to this task 100%. But can you move creation of EDgeAppManifest to a separtate function for those that use the same values please?

src/commands/mod.rs
sergey-borovkov: I see, ok that's fine for now.

src/commands/mod.rs
```diff
@@ -680,4 +909,93 @@ settings:
...
+
+        write_to_tempfile(&dir, file_name, content);
+        let file_path = dir.path().join(file_name);        
+        assert_eq!(EdgeAppManifest::validate_file(&file_path).unwrap(), false);
+    }
+
+    #[test]
+    fn test_validate_file_when_empty_field_should_return_false() {
+        let dir = tempdir().unwrap();
+        let file_name = "test.yaml";
```
sergey-borovkov, line 964: it's screenly.yml

StanIwan: It doesn't matter for tests but I can change it

src/commands/mod.rs
```diff
@@ -275,6 +315,43 @@ where
...
+
+fn deserialize_setting_type<'de, D>(deserializer: D) -> Result<SettingType, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: String = Deserialize::deserialize(deserializer)?;
+    match s.to_lowercase().as_str() {
+        "string" => Ok(SettingType::String),
+        "secret" => Ok(SettingType::Secret),
+        _ => Err(serde::de::Error::custom("Field 'type_' must be either String or Secret")),
```
sergey-borovkov, line 351: it should be field type...

src/commands/mod.rs
```diff
@@ -288,6 +365,16 @@ impl EdgeAppManifest {
         write!(&manifest_file, "---\n{yaml}")?;
         Ok(())
     }
+
+    pub fn validate_file(path: &Path) -> Result<bool, CommandError> {
+        match serde_yaml::from_str::<EdgeAppManifest>(&fs::read_to_string(path)?) {
```
sergey-borovkov, line 370: we already have code that constructs Manifest. It can be reused.

src/commands/mod.rs
```diff
@@ -275,6 +315,43 @@ where
...
+        _ => Ok(opt),
+    }
+}
+
+
+fn serialize_setting_type<S>(setting_type: &SettingType, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    let s = match setting_type {
```
sergey-borovkov, line 336: just call to_string(0) (Implement display trait or auto derive it and then convert to lower case)

src/commands/mod.rs
sergey-borovkov: Also please instead implement from str on the enum

StanIwan: "Field type must be either String or Secret"?

src/cli.rs
```diff
@@ -356,6 +356,12 @@ pub enum EdgeAppCommands {
         #[arg(short, long)]
         app_id: Option<String>,
     },
+    /// Validates screenly.yml manifest file
```
sergey-borovkov, line 359: Validates Edge App manifest file

src/cli.rs
```diff
@@ -1130,6 +1136,22 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
...
             edge_app_command.generate_mock_data(&manifest_path).unwrap();
+        },
+        EdgeAppCommands::Validate { path } => {
+            let manifest_path = transform_edge_app_path_to_manifest(path);
+            match EdgeAppManifest::validate_file(&manifest_path) {
+                Ok(true) => {
+                    println!("Manifest file is valid.");
+                },
+                Ok(false) => {
+                    println!("Manifest file is invalid.");
```
sergey-borovkov, line 1147: This is not good. More details... What is user supposed to do with this info?? You are not printing the actual error...

sergey-borovkov: Why is it even returning false? If it's not validated it should error. It should be called ensure_manifest_is_validated or something

src/commands/mod.rs
sergey-borovkov: Field type should be one of the following:
- xxx
- xxx

And for that you ideally enumerate all enum values. It's ok if you don't do that programmatically. But it should be simple and won't require changes in any extra places if we add new setting type.

sergey-borovkov: This can be done with macros of some crate like https://docs.rs/strum/latest/strum/

src/cli.rs
StanIwan: Printing error is inside the validate_file - I'll adjust it

sergey-borovkov: But why can your function still return an error then? Makes no sense to me tbh. If you already handle it inside your function

src/cli.rs
```diff
@@ -1130,6 +1136,18 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
...
             edge_app_command.generate_mock_data(&manifest_path).unwrap();
+        },
+        EdgeAppCommands::Validate { path } => {
+            let manifest_path = transform_edge_app_path_to_manifest(path);
+            match EdgeAppManifest::ensure_manifest_is_validated(&manifest_path) {
+                Ok(()) => {
+                    println!("Manifest file is valid.");
+                },
+                Err(e) => {
+                    println!("{}", e);
```
sergey-borovkov, line 1147: you can put e inside the formatting braces

sergey-borovkov (changes requested):
There are no checks for app_id not being empty everywhere.

src/commands/mod.rs
```diff
@@ -233,11 +267,24 @@ pub struct EdgeAppManifest {
     pub settings: Vec<Setting>,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Default, EnumString, Display, EnumIter)]
+pub enum SettingType {
+    #[default]
+    #[strum(serialize = "String", to_string = "string")]
```
sergey-borovkov, line 273: why is it starting with capital letter???

sergey-borovkov:
> 1. needs unit tests
> 2. needs to handle values being none in the payload
> 3. there is no validate command
> 4. needs to check that app id passed to functions is not empty or show an error in that case

Point 2 is not implemented. Need unit tests for how payload is treated when it's None.
Point 4 is not implemented at all.

src/commands/mod.rs
StanIwan: to represent enum value name as string 1:1 but I can just make it #[strum(serialize = "string")] and it will work

sergey-borovkov: Why do you want to represent it 1:1? Our spec says string/secret. not String/Secret. Please don't change stuff like that.

StanIwan:
> > 1. needs unit tests
> > 2. needs to handle values being none in the payload
> > 3. there is no validate command
> > 4. needs to check that app id passed to functions is not empty or show an error in that case
> 
> Point 2 is not implemented. Need unit tests for how payload is treated when it's None. Point 4 is not implemented at all.

Point 4 adressed with this commit -> https://github.com/Screenly/cli/pull/96/commits/a778be650d71db2321c75bd354b236e733c2d4d8

StanIwan:
> > 1. needs unit tests
> > 2. needs to handle values being none in the payload
> > 3. there is no validate command
> > 4. needs to check that app id passed to functions is not empty or show an error in that case
> 
> Point 2 is not implemented. Need unit tests for how payload is treated when it's None. Point 4 is not implemented at all.

Point 2 addressed with -> https://github.com/Screenly/cli/pull/96/commits/6becaac38607dff79a593d2432c25126dcfb7718

src/commands/mod.rs
```diff
@@ -288,6 +378,29 @@ impl EdgeAppManifest {
         write!(&manifest_file, "---\n{yaml}")?;
         Ok(())
     }
+
+    pub fn prepare_payload(manifest: &EdgeAppManifest) -> HashMap<&str, serde_json::Value> {
```
sergey-borovkov, line 382: This is not public function

src/cli.rs
```diff
@@ -508,6 +514,9 @@ fn get_actual_app_id(
     path: &Option<String>,
 ) -> Result<String, CommandError> {
     match app_id {
+        Some(id) if id.is_empty() => {
+            Err(CommandError::EmptyAppId("this".to_string()))
```
sergey-borovkov, line 518: what's "this?"?

src/commands/mod.rs
```diff
@@ -737,4 +1066,53 @@ settings:
 "#
         );
     }
+
+    #[test]
+    fn test_prepare_manifest_payload_includes_some_fields() {
```
sergey-borovkov, line 1071: No. This approach is too low level. As I mentioned it should be tested in the test_upload. You should check what actual payload ends up being.

src/cli.rs
sergey-borovkov: Hmm. Seems to be not very obviuous. So the only command that actually needs it is upload.

Because we should not pass empty app id to function. Rather when handling CLI commands we should show an early error.

Like I mentioned to you before you need to modify get_actual_app_id to return Error. It should return an error when app_id ends up being empty.

The only one that has to be outside of this is upload because it reads the whole manifest (we could move reading manifest to CLI handling but I am ok with it staying as it is for now)

StanIwan: > Hmm. Seems to be not very obviuous. So the only command that actually needs it is upload.
> 
> Because we should not pass empty app id to function. Rather when handling CLI commands we should show an early error.
> 
> Like I mentioned to you before you need to modify get_actual_app_id to return Error. It should return an error when app_id ends up being empty.

Well, that's exactly what I did - first case in match throws error if the `app_id` provided by the user is empty, then if `app_id` is empty in the manifest file, a separate error is being thrown. 

> 
> The only one that has to be outside of this is upload because it reads the whole manifest (we could move reading manifest to CLI handling but I am ok with it staying as it is for now)

sergey-borovkov: @StanIwan sorry I can't find it for other commands like promote - can you show me where they handle that?

sergey-borovkov: Ok I see. Does it work properly? Did you test it with all commands?

sergey-borovkov: Also you can remove "this" string. I don't see the point - we always know which command we call. You can always match if it's a MissingAppId error and show that.

StanIwan: I tested it with some, can test it with all

src/commands/mod.rs
StanIwan: Addressed with -> https://github.com/Screenly/cli/pull/96/commits/10ec4da1b1d7a81371fe780eba68dc2c2784ea82

src/cli.rs
StanIwan: Now, I tested it with most commands and works the same every time. Also, updated error string.

src/cli.rs
```diff
@@ -1130,6 +1139,18 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
...
             edge_app_command.generate_mock_data(&manifest_path).unwrap();
+        },
+        EdgeAppCommands::Validate { path } => {
+            let manifest_path = transform_edge_app_path_to_manifest(path);
+            match EdgeAppManifest::ensure_manifest_is_validated(&manifest_path) {
+                Ok(()) => {
+                    println!("Manifest file is valid.");
+                },
+                Err(e) => {
+                    println!("{e}");
```
sergey-borovkov, line 1150: eprintln for errors.

sergey-borovkov (approved):

src/cli.rs
```diff
@@ -508,13 +514,16 @@ fn get_actual_app_id(
...
+        },
         Some(id) => Ok(id.clone()),
         None => {
             let manifest =
                 EdgeAppManifest::new(transform_edge_app_path_to_manifest(path).as_path()).unwrap();
             match manifest.app_id {
-                Some(id) => Ok(id),
-                None => {
+                Some(id) if !id.is_empty() => Ok(id.clone()),
+                _ =>  {
```
korvyashka, line 526: Minor: None is ok here and more readable - what else could it be?)

src/commands/edge_app.rs
```diff
@@ -954,6 +950,18 @@ mod tests {
     use crate::commands::edge_app_server::MOCK_DATA_FILENAME;
     use tempfile::tempdir;
 
+    fn create_edge_app_manifest(settings: Vec<Setting>) -> EdgeAppManifest {
```
korvyashka, line 953: Minor: Let's add smth to a name to point out it is specific to test.

src/commands/mod.rs
```diff
@@ -7,9 +7,13 @@ use std::fs;
 use std::fs::File;
 use std::io::Write;
 use std::path::Path;
+use std::str::FromStr;
 use std::time::Duration;
 
 use serde::{Deserialize, Deserializer, Serialize};
+use serde_json::json;
+use strum::IntoEnumIterator;
```
korvyashka, line 15: I don't see this one used.

src/commands/mod.rs
```diff
@@ -275,6 +325,46 @@ where
...
+    let valid_setting_types = SettingType::iter()
+        .map(|t| t.to_string())
+        .collect::<Vec<_>>()
+        .join("\n") + "\n";
+
+    match SettingType::from_str(&s.to_lowercase()) {
+        Ok(setting_type) => Ok(setting_type),
+        Err(_) => Err(serde::de::Error::custom(format!(
+            "Field type should be one of the following:\n{}",
+            valid_setting_types
```
korvyashka, line 363: Here it seems it will print String\nSecret instead of string\nsecret

src/commands/mod.rs
```diff
@@ -659,16 +772,150 @@ mod tests {
...
+    title: username
+    optional: true
+    help_text: An example of a setting that is used in index.html
+"#;
+
+        assert_eq!(contents, expected_contents);
+    }
+
+    #[test]
+    fn test_save_to_file_should_skip_none_optional_fields() {
```
korvyashka, line 826: save_manifest_to...

korvyashka (approved):

src/cli.rs
StanIwan: It can be Some(id) and empty :)

src/commands/mod.rs
StanIwan: `use serde_json::json;` -> is used for ```.filter_map(|(key, value)| value.as_ref().map(|v| (*key, json!(v))))``` to omit None fields in the payload
`use strum::IntoEnumIterator;` -> is used for ```let valid_setting_types = SettingType::iter()```

src/commands/mod.rs
StanIwan: It fortunately won't because strum serialization is defined with lowercase 
```
pub enum SettingType {
    #[default]
    #[strum(serialize = "string")]
    String,
    #[strum(serialize = "secret")]
    Secret,
}
```

StanIwan: image

PR #107, opened 2023-09-11
Fix validation issues

StanIwan:
## What does this PR do?
Fixes validation issues:
- Now, we check if manifest is valid before we try to get it (commands no longer panic on invalid manifest)
- Since code around `EdgeAppManifest` and `Settings` bloated, I moved them to separate files.
- Field name is now visible in the error message:
image

- For `Option<String>` fields can easily modify if they should throw error on field being in the manifest file but empty 

## How has this been tested?
- Manually + unit tests
image

## Checklist before merging

- [x] If have added tests.
- [x] I've attached relevant screenshots (if relevant).


StanIwan:
@sergey-borovkov, @vpetersson  after this changes, if we want to allow empty string for `Option<String>` field it's as simple as changing the true->false

Of of the following optional fields, which do we allow as empty strings?
- user_version
- description
- icon
- author
- homepage_url

@sergey-borovkov mentioned that we would like to have stricter validation for fields in the `Setting` struct (e. g. fail if `help_text` is empty) - is this the only field that requires it?

StanIwan:
@sergey-borovkov When `edge-app` commands fail, some use `eprintln!`, some use `error!` - I think we should make it uniform, which way should we go with?

image

sergey-borovkov:
Yes, regarding help_text.
All optional fields in Manifest itself we decided to allow as empty. But make sure they don't serialize to xxx: null.

PR #109, opened 2023-09-11
Fix: copy_assets to version to be called on all files before upload.

korvyashka:
Refs https://phorge.wireload.net/T7493

According to description from the task above, cli call is updated to pass all files signature and upload only ones, that was not copied on the server(returned as response from the call).

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -339,7 +339,7 @@ impl EdgeAppCommand {
         let revision = self.get_latest_revision(actual_app_id).unwrap_or(0);
 
         let remote_files = self.get_version_asset_signatures(actual_app_id, revision)?;
-        let changed_files = detect_changed_files(&local_files, &remote_files)?;
+        let mut changed_files = detect_changed_files(&local_files, &remote_files)?;
         debug!("Changed files: {:?}", &changed_files);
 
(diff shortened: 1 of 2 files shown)


src/commands/edge_app.rs
```diff
@@ -339,7 +339,7 @@ impl EdgeAppCommand {
         let revision = self.get_latest_revision(actual_app_id).unwrap_or(0);
 
         let remote_files = self.get_version_asset_signatures(actual_app_id, revision)?;
-        let changed_files = detect_changed_files(&local_files, &remote_files)?;
+        let mut changed_files = detect_changed_files(&local_files, &remote_files)?;
```
sergey-borovkov, line 342: Does it have to be mut ? I can't understand from name why it has to be and other parameter is Const. can you return the new updated list instead ?

sergey-borovkov (changes requested):

src/commands/edge_app.rs
korvyashka: There is a new property copied in the FileChanges, that is to be updated later after the call to copy_assets. I can move it out from FileChanges if you want to make it cleaner.

sergey-borovkov: I think it does so if you could and it does not take too much effort please do

sergey-borovkov (approved):

PR #114, opened 2023-09-13
Updates: Deletion of edge-app to be async - call PATCH delete=true.

korvyashka:
Refs https://phorge.wireload.net/T7430

Call PATCH delete=true instead of DELETE, cause it is now async deletion, DELETE is prohibited.


sergey-borovkov (approved):
Please update the message when the command executes. Mention that delete was scheduled and it will be removed in the immediate future (I suggest using chat gpt for nice sounding message)

PR #115, opened 2023-09-14
Updates: message for deleted edge-app

korvyashka:
Refs https://phorge.wireload.net/T7430

Updates message for cli when Edge App is deleted, that it is not instant.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1094,7 +1094,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
 
             match edge_app_command.delete_app(&actual_app_id) {
                 Ok(()) => {
-                    info!("App deleted successfully.");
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -1094,7 +1094,7 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
 
             match edge_app_command.delete_app(&actual_app_id) {
                 Ok(()) => {
-                    info!("App deleted successfully.");
+                    info!("Edge App Deletion in Progress.\nRequest to delete the Edge App has been received and is now being processed. The deletion is marked for asynchronous handling, so it won't happen instantly.");
```
sergey-borovkov, line 1097: @korvyashka  - don't use "info!" for user-facing messages. They will have a lot of technical data in the formatting list we don't want to show to users.

sergey-borovkov (approved):

PR #116, opened 2023-09-14
Fixed: Array values are wrapped with double quotes

MuratUrsavas:
## What does this PR do?
Fixes wrapping array values with extra double quotes.

https://phorge.wireload.net/T7534

## Checklist before merging

- [x] If have added tests.

### DIFF
--- a/src/commands/edge_app_server.rs
+++ b/src/commands/edge_app_server.rs
@@ -1,7 +1,7 @@
 use crate::commands::ignorer::Ignorer;
 use anyhow::Result;
 use futures::future::{self, BoxFuture, FutureExt};
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::fs;
 
 use serde::{Deserialize, Serialize};
@@ -13,10 +13,21 @@ use warp::{Filter, Rejection, Reply};
 
 pub const MOCK_DATA_FILENAME: &str = "mock-data.yml";
 
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub enum Value {
+    Str(String),
+    Array(Vec<String>),
+}
+
 pub async fn run_server(
     path: &Path,
     secrets: Vec<(String, String)>,
 ) -> Result<String, anyhow::Error> {
+    let secrets_val = secrets
+        .iter()
+        .map(|(k, v)| (k.clone(), Value::Str(v.clone())))
+        .collect::<Vec<(_, _)>>();
+
     let dir_path = Arc::new(path.to_path_buf());
 
     let ignorer = Arc::new(Ignorer::new(&*dir_path)?);
@@ -34,7 +45,7 @@ pub async fn run_server(
             },
         );
 
-    let secrets_map: HashMap<_, _> = secrets.into_iter().collect();
+    let secrets_map: Vec<(_, _)> = secrets_val.into_iter().collect();
     let secrets_clone = secrets_map;
 
     let virtual_file = warp::path("edge")
@@ -74,9 +85,10 @@ struct WarpError(anyhow::Error);
 
 impl Reject for WarpError {}
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Debug, Deserialize, Serialize, Clone)]
 pub struct Metadata {
     coordinates: Vec<String>,
+    hardware: String,
     hostname: String,
     location: String,
     screen_name: String,
@@ -86,6 +98,7 @@ impl Default for Metadata {
     fn default() -> Self {
         Metadata {
             coordinates: vec!["37.3861".to_string(), "-122.0839".to_string()],
+            hardware: "x86".to_string(),
             hostname: "srly-t6kb0ta1jrd9o0w".to_string(),
             location: "Silicon Valley, USA".to_string(),
             screen_name: "Code Cafe Display".to_string(),
@@ -97,12 +110,47 @@ impl Default for Metadata {
 #[derive(Debug, Default, Deserialize)]
 struct MockData {
     metadata: Metadata,
-    settings: HashMap<String, String>,
+    settings: Vec<(String, Value)>,
+}
+
+#[derive(Debug, Default, Deserialize, Clone)]
+struct MockDataStr {
+    metadata: Metadata,
+    settings: BTreeMap<String, String>,
+}
+
+impl MockData {
+    fn new_from_str(mock_data_str: &MockDataStr) -> Self {
+        let settings_val = mock_data_str
+            .settings
+            .iter()
+            .map(|(k, v)| {
+                if v.len() > 2
+                    && v.chars().nth(0) == Some('[')
+                    && v.chars().nth(v.len() - 1) == Some(']')
+                {
+                    let v = &v[1..(v.len() - 1)];
+                    let v = v
+                        .split(',')
+                        .map(|s| s.trim().replace(['"', '\''], ""))
+                        .collect::<Vec<_>>();
+                    (k.clone(), Value::Array(v.clone()))
+                } else {
+                    (k.clone(), Value::Str(v.clone()))
+                }
+            })
+            .collect::<Vec<(_, _)>>();
+
+        MockData {
+            metadata: mock_data_str.metadata.clone(),
+            settings: settings_val,
+        }
+    }
 }
 
 async fn generate_content(
     dir_path: Arc<PathBuf>,
-    secrets: &HashMap<String, String>,
+    secrets: &[(String, Value)],
 ) -> Result<impl Reply, Rejection> {
     let file_path = dir_path.join(MOCK_DATA_FILENAME);
 
@@ -110,68 +158,87 @@ async fn generate_content(
         fs::read_to_string(&file_path).unwrap_or("".to_string())
     } else {
         eprintln!(
-            "Mock data does not exist. Use \"edge-app generate-mock-data\" to create mock data."
+            "Mock data does not exist. Use \"edge-app run --generate-mock-data\" to create mock data."
         );
         return Err(warp::reject::not_found());
     };
 
-    let data: MockData = match serde_yaml::from_str(&content) {
-        Ok(data) => data,
+    let data_str: MockDataStr = match serde_yaml::from_str(&content) {
+        Ok(data_str) => data_str,
         Err(e) => {
-            eprintln!("Mock data deserialization Error: {:?}. Use \"edge-app generate-mock-data\" to create mock data.", e);
+            eprintln!("Mock data deserialization Error: {:?}. Use \"edge-app run --generate-mock-data\" to create mock data.", e);
             return Err(warp::reject::not_found());
         }
     };
 
+    let data = MockData::new_from_str(&data_str);
     let js_output = format_js(data, secrets);
 
     Ok(warp::reply::html(js_output))
 }
 
-fn format_js(data: MockData, secrets: &HashMap<String, String>) -> String {
+fn format_js(data: MockData, secrets: &[(String, Value)]) -> String {
     format!(
         "var screenly = {{\n{metadata},\n{settings},\n{secrets},\n{cors_proxy}\n}};",
         metadata = format_section("metadata", &hashmap_from_metadata(&data.metadata)),
         settings = format_section("settings", &data.settings),
         secrets = format_section("secrets", secrets),
-        cors_proxy = "    cors_proxy_url: 'http://127.0.0.1:8080'"
+        cors_proxy = "    cors_proxy_url: \"http://127.0.0.1:8080\""
     )
 }
 
-fn hashmap_from_metadata(metadata: &Metadata) -> HashMap<String, String> {
-    let mut result = HashMap::new();
-    result.insert(
-        "coordinates".to_string(),
-        format!(
-            "[\"{}\", \"{}\"]",
-            metadata.coordinates[0], metadata.coordinates[1]
+fn hashmap_from_metadata(metadata: &Metadata) -> Vec<(String, Value)> {
+    let result = vec![
+        (
+            "coordinates".to_string(),
+            Value::Array(metadata.coordinates.clone()),
         ),
-    );
-    result.insert("hostname".to_string(), metadata.hostname.clone());
-    result.insert("location".to_string(), metadata.location.clone());
-    result.insert("screen_name".to_string(), metadata.screen_name.clone());
-    result.insert(
-        "tags".to_string(),
-        format!(
-            "[{}]",
-            metadata
-                .tags
-                .iter()
-                .map(|tag| format!("\"{}\"", tag))
-                .collect::<Vec<String>>()
-                .join(", ")
+        (
+            "hardware".to_string(),
+            Value::Str(metadata.hardware.clone()),
         ),
-    );
+        (
+            "hostname".to_string(),
+            Value::Str(metadata.hostname.clone()),
+        ),
+        (
+            "location".to_string(),
+            Value::Str(metadata.location.clone()),
+        ),
+        (
+            "screen_name".to_string(),
+            Value::Str(metadata.screen_name.clone()),
+        ),
+        (
+            "tags".to_string(),
+            Value::Array(
+                metadata
+                    .tags
+                    .iter()
+                    .map(|tag| tag.to_string())
+                    .collect::<Vec<String>>(),
+            ),
+        ),
+    ];
     result
 }
 
-fn format_section(name: &str, items: &HashMap<String, String>) -> String {
+fn format_section(name: &str, items: &[(String, Value)]) -> String {
     let content = items
         .iter()
-        .map(|(k, v)| format!("        \"{}\": \"{}\"", k, v))
+        .map(|(k, v)| match v {
+            Value::Str(s) => format!("        \"{}\": \"{}\"", k, s),
+            Value::Array(arr) => format!(
+                "        \"{}\": [{}]",
+                k,
+                arr.iter()
+                    .map(|item| format!("\"{}\"", item))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            ),
+        })
         .collect::<Vec<_>>()
         .join(",\n");
-
     format!("    {}: {{\n{}\n    }}", name, content)
 }
 
@@ -193,13 +260,14 @@ metadata:
   coordinates:
   - '37.3861'
   - '-122.0839'
+  hardware: x86
   hostname: srly-t6kb0ta1jrd9o0w
   location: Silicon Valley, USA
   screen_name: Code Cafe Display
   tags:
   - All Screens
 settings:
-  disable_analytics: 'false'
+  enable_analytics: 'true'
   override_timezone: ''
   tag_manager_id: ''
 "#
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cli"
-version = "0.2.0"
+version = "0.2.1"
```
sergey-borovkov, line 3: Why ?

src/cli.rs
```diff
@@ -1148,9 +1148,13 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
             generate_mock_data,
         } => {
             let secrets = if let Some(secret_pairs) = secrets {
-                secret_pairs.secrets.clone()
+                secret_pairs
```
sergey-borovkov, line 1151: why does it need to be converted here? And I don't see any reason why secret pairs need to be of this type. Especially in exposed API.

Any specific reason?

src/commands/edge_app_server.rs
```diff
@@ -96,39 +102,75 @@ impl Default for Metadata {
...
+
+#[derive(Debug, Default, Deserialize, Clone)]
+struct MockDataStr {
     metadata: Metadata,
     settings: HashMap<String, String>,
 }
 
+impl MockData {
+    fn new_from_str(mock_data_str: &MockDataStr) -> Self {
+        let settings_val = mock_data_str
```
sergey-borovkov, line 117: why are we not using serde to deserialize string?

sergey-borovkov (changes requested):

sergey-borovkov:
Also there is no new test that tests for issue you fixed.

Cargo.toml
MuratUrsavas: Because the code has changed. We're using versions to tag a specific state of software. If that state changes, the version has to change.

sergey-borovkov: No. We bump version when we release new version, not for every fix (this is not even enough as the version is referenced in docker and somewhere else)

src/cli.rs
MuratUrsavas: No specific reason actually. Was just quickly following the reference chain. Maybe overdone it :smile:

src/commands/edge_app_server.rs
MuratUrsavas: Well, that was the failing part. It cannot automatically create the new `MockData` from a string. I needed to create that middle step to fix it. If you have a more idiomatic idea, I'm all ears.

MuratUrsavas:
> Also there is no new test that tests for issue you fixed.

Current tests already cover it. Maybe need to modify the current mock data a little bit.

src/commands/edge_app_server.rs
sergey-borovkov: What's the issue? It's not enought data for me that it's a failing part. But if some field fails to parse correctly you can always use custom deserializer - take a look at EdgeAppManifest for instance it does that a lot

sergey-borovkov:
> > Also there is no new test that tests for issue you fixed.
> 
> Current tests already cover it. Maybe need to modify the current mock data a little bit.

We still should cover the exact failure we had. Those tests did not reveal the issue. I'd propose to have a test that checks that the whole output is correct by comparing with the whole output string typed out in the test

MuratUrsavas:
> We still should cover the exact failure we had. Those tests did not reveal the issue. I'd propose to have a test that checks that the whole output is correct by comparing with the whole output string typed out in the test

Will do that with the mock data change.

src/cli.rs
sergey-borovkov: I feel like it's a bit more natural for us not to expose those internal types they can be private to the module so that consumers don't have to care about it

Cargo.toml
MuratUrsavas: How about using dash notation for intermediary versions under development? Like `0.2.1-1` and while releasing we can get rid of the dash suffix.

src/cli.rs
MuratUrsavas: Actually, the exposed part is still the same. Are you sure we're exposing the new type? Maybe I'm overseeing something?

src/commands/edge_app_server.rs
MuratUrsavas: I'd like to learn that custom deserializer. But we need to ship this one quickly. How about implementing it as a follow-up?

Cargo.toml
sergey-borovkov: I don't see any benefit to that as you can't download it from any place. And it's what we do in all our projects so what's the point of doing differently here?

src/cli.rs
sergey-borovkov: EdgeAppServerValue::Str is exposed which is a private enum

src/commands/edge_app_server.rs
sergey-borovkov: ok since we need it fast - please create a GH issue and reference it there

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cli"
-version = "0.2.0"
+version = "0.2.1-1"
```
sergey-borovkov, line 3: pls no :)

Cargo.toml
MuratUrsavas: We're testing it here and there and we should be providing it to other teams for tests before releasing it. And every software iteration has to carry a different tag. Reports have to include that version info to match the exact state of that software. You can use build number, commit hash, or anything else. But some info has to indicate that state. Version is the easiest one.

sergey-borovkov: When it's unreleased that's what you have git commit hash for. We are *not* providing that as a binary for testing to anyone.

src/commands/edge_app_server.rs
```diff
@@ -227,16 +285,21 @@ settings:
         assert!(content.contains("metadata: {")); // Make sure metadata is in its own section
         assert!(content.contains("\"srly-t6kb0ta1jrd9o0w\"")); // Checking an example metadata value
         assert!(content.contains("key"));
+        assert!(content.contains("[\"37.3861\", \"-122.0839\"]"));  // Make sure arrays are written correctly
```
sergey-borovkov, line 288: Let's test the whole file content, please. 

instead of contains using strict eq. 

I did this approach and it was an error since it turned out to cause bugs let's not use it further.

src/cli.rs
MuratUrsavas: Done!

Cargo.toml
MuratUrsavas: OK :)

src/commands/edge_app_server.rs
MuratUrsavas: OK, let's do this one as a follow-up. I've done it but it kept invalidating the valid string generated by itself. Will take a deeper look after that. Approve and release as it is.

src/commands/edge_app_server.rs
MuratUrsavas: Will do.

src/commands/edge_app_server.rs
```diff
@@ -214,29 +269,39 @@ settings:
...
 
+        print!("{}", content);
+
         assert!(content.contains("screenly ="));
         assert!(content.contains("http://127.0.0.1:8080")); // Assuming this is expected from your server
         assert!(content.contains("metadata: {")); // Make sure metadata is in its own section
         assert!(content.contains("\"srly-t6kb0ta1jrd9o0w\"")); // Checking an example metadata value
         assert!(content.contains("key"));
+        assert!(content.contains("[\"37.3861\", \"-122.0839\"]"));  // Make sure arrays are written correctly
+        assert!(!content.contains("\"[\"37.3861\", \"-122.0839\"]\""));  // Make sure arrays are not written incorrectly
```
sergey-borovkov, line 291: don't check for negatives, please...

src/commands/edge_app_server.rs
sergey-borovkov: I can't approve when the test for the actual content does not pass :( Maybe someone else will

src/commands/edge_app_server.rs
```diff
@@ -138,40 +180,53 @@ fn format_js(data: MockData, secrets: &HashMap<String, String>) -> String {
...
-            "[{}]",
+        Value::Array(
             metadata
                 .tags
                 .iter()
-                .map(|tag| format!("\"{}\"", tag))
-                .collect::<Vec<String>>()
-                .join(", ")
+                .map(|tag| format!("{}", tag))
+                .collect::<Vec<String>>(),
```
sergey-borovkov, line 208: Clippy warning needs to be fixed

src/commands/edge_app_server.rs
```diff
@@ -96,39 +102,75 @@ impl Default for Metadata {
...
+            .iter()
+            .map(|(k, v)| {
+                if v.len() > 2
+                    && v.chars().nth(0) == Some('[')
+                    && v.chars().nth(v.len() - 1) == Some(']')
+                {
+                    let v = &v[1..(v.len() - 1)];
+                    let v = v
+                        .split(',')
+                        .map(|s| s.trim().replace('"', "").replace('\'', ""))
```
renatgalimov, line 128:
```suggestion
                        .map(|s| s.trim().replace(['"', '\''], ""))
```

src/commands/edge_app_server.rs
```diff
@@ -138,40 +180,53 @@ fn format_js(data: MockData, secrets: &HashMap<String, String>) -> String {
...
-        format!(
-            "[{}]",
+        Value::Array(
             metadata
                 .tags
                 .iter()
-                .map(|tag| format!("\"{}\"", tag))
-                .collect::<Vec<String>>()
-                .join(", ")
+                .map(|tag| format!("{}", tag))
```
renatgalimov, line 207:
```suggestion
                .map(|tag| format!(tag.to_string()))
```

src/commands/edge_app_server.rs
sergey-borovkov: But there is no created issue and it's not referenced here :( So it's not resolved

src/commands/edge_app.rs
```diff
@@ -286,15 +288,20 @@ impl EdgeAppCommand {
...
+    ) -> Result<(), anyhow::Error> {
         let address_shared = Arc::new(Mutex::new(None));
         let address_clone = address_shared.clone();
 
         let runtime = tokio::runtime::Runtime::new().unwrap();
         let path = path.to_path_buf();
         runtime.block_on(async {
             tokio::spawn(async move {
-                let address = run_server(path.as_path(), secrets).await.unwrap();
+                let secrets_val = secrets.iter().map(|(k, v)| { (k.clone(), EdgeAppServerValue::Str(v.clone()))}).collect();
```
korvyashka, line 303: Shouldnt fmt format this line?

src/commands/edge_app_server.rs
renatgalimov: I agree with @sergey-borovkov, just do one equality test to a full file content.

src/commands/edge_app_server.rs
```diff
@@ -13,9 +13,15 @@ use warp::{Filter, Rejection, Reply};
...
+#[derive(Debug, Clone, Deserialize)]
+pub enum Value {
+    Str(String),
+    Array(Vec<String>),
+}
+
 pub async fn run_server(
     path: &Path,
-    secrets: Vec<(String, String)>,
+    secrets: Vec<(String, Value)>,
```
korvyashka, line 24: Hm - inst this strictly cli related format, may be pass it with string to parse later?

src/commands/edge_app_server.rs
sergey-borovkov: you don't need format! then though?

korvyashka (commented):
Commented from mobile, so could miss or missunderstood smth.

1) please always do fmt and fix all clippy warnings
2) All tests on pr should pass before merging - not sure if they are, just saw the concersation

src/commands/edge_app_server.rs
sergey-borovkov: yeah, I  also think it should all be in the server code.

src/commands/edge_app_server.rs
renatgalimov: You are absolutely right

src/commands/edge_app.rs
MuratUrsavas: Yeah forgot to run rustfmt while on my way out. Done.

src/commands/edge_app_server.rs
```diff
@@ -96,39 +102,75 @@ impl Default for Metadata {
...
+}
+
+#[derive(Debug, Default, Deserialize, Clone)]
+struct MockDataStr {
     metadata: Metadata,
     settings: HashMap<String, String>,
 }
 
+impl MockData {
+    fn new_from_str(mock_data_str: &MockDataStr) -> Self {
```
sergey-borovkov, line 116: this function also needs to be tested especially considering that you parse it manually. You need to test it with invalid input/edge cases.

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
MuratUrsavas: OK, how a test could fail even with its own output? Would you like me to wait to figure that bug out? I'm copying its left output to the right, and then assert it.

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
MuratUrsavas: https://github.com/Screenly/cli/issues/119 :smile:

src/commands/edge_app_server.rs
MuratUrsavas: Done.

src/commands/edge_app_server.rs
```diff
@@ -34,16 +45,16 @@ pub async fn run_server(
             },
         );
 
-    let secrets_map: HashMap<_, _> = secrets.into_iter().collect();
+    let secrets_map: BTreeMap<_, _> = secrets_val.into_iter().collect();
```
sergey-borovkov, line 48: Hmm. Maybe better to just use Vec<>? It preserves the order and we don't access the values by key/value

src/commands/edge_app_server.rs
```diff
@@ -221,12 +288,8 @@ settings:
             .await
             .unwrap();
         let content = resp.text().await.unwrap();
-
-        assert!(content.contains("screenly ="));
-        assert!(content.contains("http://127.0.0.1:8080")); // Assuming this is expected from your server
-        assert!(content.contains("metadata: {")); // Make sure metadata is in its own section
-        assert!(content.contains("\"srly-t6kb0ta1jrd9o0w\"")); // Checking an example metadata value
-        assert!(content.contains("key"));
+        let expected_content = "var screenly = {\n    metadata: {\n        \"coordinates\": [\"37.3861\", \"-122.0839\"],\n        \"hardware\": \"x86\",\n        \"hostname\": \"srly-t6kb0ta1jrd9o0w\",\n        \"location\": \"Silicon Valley, USA\",\n        \"screen_name\": \"Code Cafe Display\",\n        \"tags\": [\"All Screens\"]\n    },\n    settings: {\n        \"enable_analytics\": \"true\",\n        \"override_timezone\": \"\",\n        \"tag_manager_id\": \"\"\n    },\n    secrets: {\n        \"key\": \"value\"\n    },\n    cors_proxy_url: \"http://127.0.0.1:8080\"\n};";
```
sergey-borovkov, line 291: maybe format it as multiline string to make it more readable?

src/commands/edge_app_server.rs
```diff
@@ -259,4 +322,21 @@ settings:
 
         assert_eq!(resp.status(), 404);
     }
+
+    #[tokio::test]
+    async fn test_mock_data_from_str() {
```
sergey-borovkov, line 327: Please improve the naming - see here how we name our tests https://phorge.wireload.net/w/developer_guidelines/tests/

should contain what it tests and the expected behavior (but see the document anyway for in depth explanation and examples)

src/commands/edge_app_server.rs
MuratUrsavas: I think that could be left as it is.

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
MuratUrsavas: Done!

src/commands/edge_app_server.rs
sergey-borovkov: We are not using it as a map but OK

sergey-borovkov (approved):

src/commands/edge_app_server.rs
StanIwan: I'm with Sergey on this one, even though BTreeMap might work, if we can also just use Vec, that would be preferrable.

StanIwan (approved):
I approve, this are clippy warnings I get on 1.70:
```
siwan@siwan-x1-carbon:~/Projects/cli$ cargo clippy
warning: redundant clone
   --> src/commands/edge_app_server.rs:137:47
    |
137 |                     (k.clone(), Value::Array(v.clone()))
    |                                               ^^^^^^^^ help: remove this
    |
note: this value is dropped without further use
   --> src/commands/edge_app_server.rs:137:46
    |
137 |                     (k.clone(), Value::Array(v.clone()))
    |                                              ^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: `#[warn(clippy::redundant_clone)]` on by default

warning: called `.nth(0)` on a `std::iter::Iterator`, when `.next()` is equivalent
   --> src/commands/edge_app_server.rs:129:24
    |
129 |                     && v.chars().nth(0) == Some('[')
    |                        ^^^^^^^^^^^^^^^^ help: try calling `.next()` instead of `.nth(0)`: `v.chars().next()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#iter_nth_zero
    = note: `#[warn(clippy::iter_nth_zero)]` on by default

warning: `cli` (bin "screenly") generated 2 warnings (run `cargo clippy --fix --bin "screenly"` to apply 2 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
```

src/commands/edge_app_server.rs
MuratUrsavas: Yeah, I know but there is a great possibility we could use it soon. Also, it doesn't create any dev overhead. That's why I opted to leave as it is.

sergey-borovkov: @MuratUrsavas what possibility? I am not aware of it

Cargo.toml
renatgalimov: Guys. We use semantic versioning. We modify the version when we release. 
If you for whatever reason want to modify the version between releases you can use -rc tail.

It mains release candidate. 

2.0.0-rc1. 
2.0.0-rc2.
2.0.0-rc3.

We also can release the same version multiple times. We had to do it recently, to trigger the upgrade on UC16 devices. Then you can add -r tail

It means - revision.

2.0.0-r1 means 2.0.0 revision 1.

src/commands/edge_app_server.rs
korvyashka: BTree is usually used in the databases(indexes) for fast reads(searches) specifically, there is usually no reason using it in the day to day code, unless there are a lot of items in the memory you want to search in fast. So I don't see any pros using it here.

renatgalimov (approved):

src/commands/edge_app_server.rs
MuratUrsavas: OK, guys, you won. I've converted most of the BTreeMaps to Vec's but kept in YAML conversion. Without a custom deserializer, I cannot change it to a Vec. It will be done in another task.

vpetersson:
Woop woop!

Issue #119, opened 2023-09-21
Add custom deserializer for MockData from MockDataStr


MuratUrsavas:
Right now we're doing the conversion semi-manually. But it would be better to do it as a deserializer for serde.

sergey-borovkov:
Fixed by https://github.com/Screenly/cli/pull/121/ by removing the need to have the custom deserializer

PR #121, opened 2023-09-22
Remove custom deserializer

sergey-borovkov:
## What does this PR do?
Removes custom serializer. Simplifies data structure.

https://github.com/Screenly/cli/issues/119

## How has this been tested?
All tests pass and `edge-app run` with hello world works in the browser..
## Checklist before merging

- [ ] If have added tests.

### DIFF
--- a/src/commands/edge_app_server.rs
+++ b/src/commands/edge_app_server.rs
@@ -1,7 +1,7 @@
 use crate::commands::ignorer::Ignorer;
 use anyhow::Result;
 use futures::future::{self, BoxFuture, FutureExt};
-use std::collections::{HashMap};
+use std::collections::HashMap;
 use std::fs;
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


src/commands/edge_app_server.rs
```diff
@@ -110,42 +110,7 @@ impl Default for Metadata {
...
-            .iter()
-            .map(|(k, v)| {
-                if v.len() > 2
-                    && v.chars().nth(0) == Some('[')
-                    && v.chars().nth(v.len() - 1) == Some(']')
-                {
-                    let v = &v[1..(v.len() - 1)];
-                    let v = v
-                        .split(',')
-                        .map(|s| s.trim().replace(['"', '\''], ""))
```
MuratUrsavas, line 135: What happens if the value includes quotes? This custom deserializer was also discarding them.

korvyashka (approved):

src/commands/edge_app_server.rs
sergey-borovkov: It's a proper yaml parser. It won't bring in quotes denoting the string.

sergey-borovkov: Our test `test_server_should_serve_correct_mock_data` already tests exactly for that:

image

sergey-borovkov: Also does not that mean your code would not handle " inside the string? It's valid to have it for strings.

MuratUrsavas: If that's taken care of, I'm OK. It was an issue in the beginning I guess but I might be recalling it incorrectly.

MuratUrsavas (approved):

PR #129, opened 2023-10-23
Updates: required string can lack default_value. Updated undefined ap…

korvyashka:
Refs https://phorge.wireload.net/T7672

As we removed constraint from our server, enforsing string setting to have default values, it is also removed from cli.
Also updated url secrets/undefined to settings/undefined as it now returns all undefined settings, not only secrets.

This will require new release, when mentioned changes will be on production.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -406,9 +406,9 @@ impl EdgeAppCommand {
         revision: u32,
         channel: &String,
     ) -> Result<(), CommandError> {
-        let secrets = self.get_undefined_secrets(app_id)?;
+        let secrets = self.get_undefined_settings(app_id)?;
         if !secrets.is_empty() {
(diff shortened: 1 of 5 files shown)


src/commands/edge_app.rs
```diff
@@ -554,7 +554,7 @@ impl EdgeAppCommand {
         let undefined_secrets_response = commands::get(
```
sergey-borovkov, line 554: unddefined_settings then?

src/commands/edge_app.rs
```diff
@@ -2007,7 +2007,7 @@ mod tests {
         //  v4/edge-apps/settings?select=type,default_value,optional,title,help_text&app_id=eq.{}&order=title.asc
         let undefined_secrets_mock = mock_server.mock(|when, then| {
```
sergey-borovkov, line 2008: same

sergey-borovkov (approved):

PR #140, opened 2023-11-29
Fix nix builds on macOS

jnsgruk:
## What does this PR do?

The build for macOS was subtly broken due to some missing frameworks. This PR addresses that and makes the build pass, but also adds a CI job to run the build on macOS alongside Ubuntu. I've also made the CI workflow run on a few more paths.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,40 +22,44 @@
       });
     in
     {
-      overlays.default = final: _prev: let 
-        inherit ((builtins.fromTOML(builtins.readFile ./Cargo.toml)).package) version;
-        inherit (final) openssl perl pkg-config stdenv lib darwin rustPlatform;
-      in {
-        screenly-cli = rustPlatform.buildRustPackage rec {
-          inherit version;
-          pname = "screenly-cli";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -22,40 +22,44 @@
...
-            openssl
-          ] ++ lib.optionals stdenv.isDarwin [
-            darwin.apple_sdk.frameworks.CoreFoundation
-            darwin.apple_sdk.frameworks.Security
-          ];
+            buildInputs = [
+              openssl
+            ] ++ lib.optionals stdenv.isDarwin [
+              darwin.apple_sdk.frameworks.CoreFoundation
+              darwin.apple_sdk.frameworks.CoreServices
```
jnsgruk, line 48: This was missing before

vpetersson (approved):

flake.nix
```diff
@@ -22,40 +22,44 @@
...
+              darwin.apple_sdk.frameworks.Security
+              darwin.apple_sdk.frameworks.SystemConfiguration
```
jnsgruk, line 50: This was missing before

PR #150, opened 2024-02-28
Adds: detecting version metadata changes - consider changed as a new …

korvyashka:
Refs https://phorge.wireload.net/T8225

Problem before is that just changing icon or description doesn't trigger new version to be created.
We also can't change a version once it is published - it is set in stone.

So now we detect when version metadata is changed and consider it a new version.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -974,12 +974,12 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
                         Ok(rev) => match rev {
                             Some(rev) => rev.revision,
                             None => {
-                                println!("Failed to get latest edge app revision.");
+                                println!("Failed to get latest Edge App revision.");
                                 std::process::exit(1);
                             }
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -971,7 +971,13 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
 
                 let revision = if *latest {
                     match edge_app_command.get_latest_revision(&actual_app_id) {
-                        Ok(rev) => rev,
+                        Ok(rev) => match rev {
+                            Some(rev) => rev.revision,
+                            None => {
+                                println!("Failed to get latest edge app revision.");
```
sergey-borovkov, line 977: Edge App :) Please fix anything related to this with regex please as Viktor asked for that speciffically

sergey-borovkov (approved):

rusko124 (approved):

PR #155, opened 2024-03-07
Adds: 4.1 API update for title and name for settings.

korvyashka:
Refs https://phorge.wireload.net/T8231

Uses 4.1 API version, where we differentiated name and title. Basically what was previously an id(named title) is now a 'name'. Though it is not affecting yaml at all, cause title was ignored there totally before.

### DIFF
--- a/src/commands/edge_app_settings.rs
+++ b/src/commands/edge_app_settings.rs
@@ -96,25 +96,29 @@ where
             for (key, value) in setting_data {
                 match key.as_str() {
                     "type" => {
-                        setting.type_ = deserialize_setting_type(value).unwrap();
+                        setting.type_ =
+                            deserialize_setting_type(value).expect("Failed to parse setting type.");
                     }
                     "default_value" => {
                         setting.default_value = value.as_str().map(|s| s.to_string());
                     }
                     "title" => {
-                        setting.title = value.as_str().unwrap().to_string();
+                        setting.title = value.as_str().expect("Failed to parse title.").to_string();
                     }
                     "optional" => {
-                        setting.optional = value.as_bool().unwrap();
+                        setting.optional = value.as_bool().expect("Failed to parse optional.")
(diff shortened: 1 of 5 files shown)


src/commands/edge_app_settings.rs
```diff
@@ -72,12 +75,55 @@ where
...
+            for (key, value) in setting_data {
+                match key.as_str() {
+                    "type" => {
+                        setting.type_ = deserialize_setting_type(value).unwrap();
+                    }
+                    "default_value" => {
+                        setting.default_value = value.as_str().map(|s| s.to_string());
+                    }
+                    "title" => {
+                        setting.title = value.as_str().unwrap().to_string();
```
sergey-borovkov, line 105: instead of unwrap do expect at least and write a human readable message please

src/commands/edge_app_utils.rs
```diff
@@ -191,17 +191,19 @@ mod tests {
...
                     type_: SettingType::String,
                     default_value: Some("5".to_string()),
-                    title: "display_time".to_string(),
+                    title: "display time title".to_string(),
                     optional: true,
                     is_global: false,
                     help_text: "For how long to display the map overlay every time the rover has moved to a new position.".to_string(),
                 },
                 Setting {
+                    name: "google_maps_api_key".to_string(),
```
sergey-borovkov, line 203: this is getting confusing though with name and title. I wish we just had a field named `id` or something. Or "key" since it's a setting. key=val. Oh well. We could rename it here and pass the correct data to API but not sure if it's worth it

korvyashka: Yeah - thought about that, but decided we could remember the name and title difference.
Also in the cli there is no name in the yaml anyway - I skip serialize and deserialize for the name

sergey-borovkov (approved):

PR #160, opened 2024-04-06
Feature/installation id manifest

korvyashka:
Bringing installation_id back, cause now we could have multiple installation in the UI.

For compatability with old yamls without installation_id, cli, when needed, will request(or create) and save installation_id using deprecated name 'Edge app cli installation'.
So eventually yaml file must have installation_id to work properly.

Other changes
- Installation of an app moved from upload to edge app create command.
- For some commands now installation_id is required either by passing as argument or in the manifest.
  - Promote command(to check undefined settings)
  - All settings commands
  - Secret set command

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -386,10 +386,6 @@ pub enum EdgeAppVersionCommands {
         #[arg(short, long, default_value = "stable")]
         channel: String,
 
-        /// Edge app id. If not specified CLI will use the id from the manifest.
-        #[arg(short, long)]
-        app_id: Option<String>,
-
         /// Edge App Installation id. If app_id is specified, installation_id must be also specified. If both are not specified, CLI will use the installation_id from the manifest
         #[arg(short, long)]
         installation_id: Option<String>,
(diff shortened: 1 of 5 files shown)


korvyashka:
TODO: update commands to have only installation_id passed as argument when both are required.

src/cli.rs
```diff
@@ -957,20 +925,31 @@ pub fn handle_cli_edge_app_command(command: &EdgeAppCommands) {
...
             } => {
-                let actual_app_id = match get_actual_app_id(app_id, path) {
-                    Ok(id) => id,
+                let actual_installation_id = match edge_app_command
+                    .ensure_installation_id(installation_id.clone(), path.clone())
+                {
+                    Ok(_installation_id) => _installation_id,
                     Err(e) => {
-                        error!("Failed to promote edge app version: {}", e);
+                        error!("Error calling promote: {}", e);
```
sergey-borovkov, line 937: "Promote failed" maybe?

sergey-borovkov (approved):

PR #169, opened 2024-05-16
Try: fixing error in ci

korvyashka:
Testing fix to linter


rusko124:
Ref: https://github.com/Screenly/cli/pull/170

PR #170, opened 2024-07-02
Edits: 'edge-app run' open in browser

rusko124:
Ticket: https://phorge.wireload.net/T8681

Edits:
- `screenly edge-app run` opens the URL in browser

It is tested only on Linux.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -391,6 +391,28 @@ impl EdgeAppCommand {
                 address_shared.lock().unwrap().as_ref().unwrap()
             );
 
+            #[cfg(target_os = "macos")]
+            {
+                let _ = std::process::Command::new("open")
+                    .arg(format!("{}/index.html", address_shared.lock().unwrap().as_ref().unwrap()))
+                    .output();
+            }
+
+            #[cfg(target_os = "linux")]
+            {
+                let _ = std::process::Command::new("xdg-open")
+                    .arg(format!("{}/index.html", address_shared.lock().unwrap().as_ref().unwrap()))
+                    .output();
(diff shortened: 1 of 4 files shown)


src/commands/edge_app.rs
```diff
@@ -391,6 +391,28 @@ impl EdgeAppCommand {
                 address_shared.lock().unwrap().as_ref().unwrap()
             );
 
+            #[cfg(target_os = "macos")]
```
renatgalimov, line 394: The whole code within the block_on block looks super lowbrow. What do you think about refactoring it like this:

1. Use CondVar to wait for the address. Wait with a timeout, raising an error if the app doesn't start in time. https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html
2. Once the shared address is available, lock and clone it immediately so you don't need to do it later. 
3. You're not analyzing the result of the command. Do it. Right after you get the output, check that it's successful, and if not - print the error.

src/commands/edge_app.rs
```diff
@@ -399,6 +405,32 @@ impl EdgeAppCommand {
         Ok(())
     }
 
+    fn open_browser(&self, address: &str) -> Result<(), CommandError> {
+        let command = match std::env::consts::OS {
+            "macos" => "open",
+            "windows" => "start",
+            "linux" => "xdg-open",
```
renatgalimov, line 412: Delightful🤩

renatgalimov (approved):

korvyashka (approved):
Please add tests if possible, though I understand it could be complicated, so approving

sergey-borovkov (approved):

PR #174, opened 2024-07-04
Clarify how to update the secret value

renatgalimov:
My intuition tells me I must set the secret value in the config file, but that's wrong. 

In the help text, I clarify how to instantiate the secret value.

## Checklist before merging

- [ ] If have added tests.

### DIFF
--- a/src/commands/edge_app.rs
+++ b/src/commands/edge_app.rs
@@ -111,7 +111,7 @@ impl EdgeAppCommand {
                     default_value: None,
                     optional: true,
                     is_global: false,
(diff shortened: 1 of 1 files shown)


korvyashka:
It is good to add in the documentation, though I don't think secret would be ever set in the config - also same applies to the setting - and we want to merge them as well - it all will be just setting command.

rusko124:
@renatgalimov I don't think the PR is still relevant with the changes in dev-branch

vpetersson (approved):

PR #176, opened 2024-07-04
New: generated help markdown documentation

rusko124:
This PR allows us to generate the full help of our CLI in markdown format. I also added a workflow job verifying that `docs/CommandLineHelp.md` has been updated. This allows us to see how new changes (PRs) affect the documentation. 
Example of the failed job: https://github.com/Screenly/cli/actions/runs/9792520351/job/27038547184

I'd like to link to this file in our articles instead of hardcoded commands.

For the doc file generation, I use [clap-markdown](https://github.com/ConnorGray/clap-markdown).


MuratUrsavas:
@rusko124 Why not make it a build task, instead of a runtime parameter? And it would be uploaded/handled by CI with the releases. With runtime parameter, I first thought it will be served directly to customer. But I guess it's not, right?

MuratUrsavas:
And it would be great if you can squash the branch.

rusko124:
> @rusko124 Why not make it a build task, instead of a runtime parameter? And it would be uploaded/handled by CI with the releases. With runtime parameter, I first thought it will be served directly to customer. But I guess it's not, right?

> Why not make it a build task
It pulls a ton of dependencies to build-dependencies. Just don't want to pollute it.

> And it would be uploaded/handled by CI with the releases. 
The file should be generated manually, and be part of the next Pull Requests. We should see what changed.

It's runtime yes, but hidden. So basically how it's done in the lib docs.

MuratUrsavas:
IMHO, this looks not right. A change should be happening atomically, and docs should reflect the same change moment. That's why the same PR should also include doc changes, and it's easy with a build task. Sure, it could be handled with a runtime param, but it doesn't feel right, even if it is hidden.

I wonder what others are thinking.

rusko124:
> IMHO, this doesn't look right. A change should be happening atomically, and docs should reflect the same change moment. That's why the same PR should also include doc changes, and it's easy with a build task. Sure, it could be handled with a runtime param, but it doesn't feel right, even if it is hidden.

Do you mean the build process should overwrite files outside of the build dir? If so it doesn't sound correct. Or, do you mean to copy a file from the build dir and replace it in the docs? It also doesn't sound more automatic than it is now.
Also during CI, we should not deliver generated documentation. It should be done manually in PRs, in CI we can only make sure that the documentation is up to date.

MuratUrsavas:
> Do you mean the build process should overwrite files outside of the build dir?

Hmmm, I think I haven't thought it thoroughly. But I guess it would be the same, as where we put them right now, in the PR.

It's just not felt right, but apparently I don't know the better one, too ;) So I withdraw.

src/cli.rs
```diff
@@ -63,6 +63,9 @@ pub enum Commands {
     /// Edge App related commands.
     #[command(subcommand)]
     EdgeApp(EdgeAppCommands),
+    // For generating `docs/CommandLineHelp.md`.
```
sergey-borovkov, line 66: well it should still be ///.. Though I guess it's hidden anyhow... So might not matter

sergey-borovkov (approved):

src/cli.rs
rusko124, line 66:
```suggestion
    /// For generating `docs/CommandLineHelp.md`.
```

PR #181, opened 2024-07-06
Feature/instance commands instance manifest

korvyashka:
Updates instance create command to work with instance manifest.
Uncommented list as it just works. update/delete test also uncommented as they work with arbitrary instance_id

Though have some open questions we need to discuss:

### Create command

Now behaviour is - we only create instance if there is no instance.yml present.
Though we could support --force to re-create instance and re-write instance.yml or --new to create new instance w/o touch instance.yml

### Update command.
We have instance.yml and not sure what to do if we update name or entrypoint in the yml file - when and how should we update that instance then?
Also update command by itself could rename the instance(or change the entrypoint I guess if implemented) - so question is - should we also update instance.yml if update is called directly?

### Delete command
Similar question is here - though I believe if we delete instance that is in the instance.yml - we just delete the file.
Should we prompt user with an instance name to confirm deletion?

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -13,7 +13,9 @@ use crate::authentication::{verify_and_store_token, Authentication, Authenticati
 use crate::commands;
 use crate::commands::edge_app_manifest::EdgeAppManifest;
 use crate::commands::edge_app_server::MOCK_DATA_FILENAME;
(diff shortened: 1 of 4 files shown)


renatgalimov:
Could you please fix Clippy warnings?

PR #262, opened 2025-09-29
Adds documentation for help_text - input field types

iamplugged:
Adds documentation for the below fields:

    Select
    Date Time
    Switch
    Number
    URL
    Text Area

### DIFF
--- a/docs/EdgeApps.md
+++ b/docs/EdgeApps.md
@@ -416,6 +416,140 @@ settings:
     help_text: An example of an API key
 ```
 
+#### Input field types
+
+Edge App settings support additional input field types beyond plain text and password. To render a specific input type in the UI, embed a small JSON descriptor inside the setting's `help_text` field. This JSON does not change the underlying storage type (which remains `string` or `secret`); it only instructs the UI how to render the input.
+
+- **Schema**: The JSON must include `schema_version` and a `properties` object.
+- **Common keys**:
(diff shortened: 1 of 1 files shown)


vpetersson (approved):
LGTM but needs to also be updated in the [developer portal](https://github.com/Screenly/developer-portal)

docs/EdgeApps.md
```diff
@@ -416,6 +416,140 @@ settings:
...
+    title: Website URL
+    optional: false
+    help_text: |
+      {
+        "schema_version": 1,
+        "properties": {
+          "help_text": "The URL of the website",
+          "type": "url"
+        }
+      }
```
renatgalimov, line 546: I find it pretty inconvenient. 
Could we deal with the type of the `help_text` internally, and in allow the data to be a dict in yaml? 
```suggestion
    help_text: 
      schema_version: 1,
      properties: 
        help_text": "The URL of the website",
        type: "url"
```

vpetersson: @renatgalimov this is indeed more native. I don't live the idea of mixing YAML and JSON either. Agree that this is a better approach.

rusko124: @iamplugged let's land this PR and create a ticket to change help_text schema from JSON to YAML

iamplugged: https://phorge.wireload.net/T9633

PR #293, opened 2026-06-06
Add multi-profile authentication support

514sid:
## Summary

Replaces the single-token `~/.screenly` file with a YAML-based profile store,
enabling multiple named authentication profiles with a single active profile at a time.

- `screenly login --name <profile>` stores a token under a named profile and switches to it.
  Without `--name` it updates the currently active profile, which is the re-login-after-token-rotation
  flow, and defaults to `"default"` on a fresh install.
- `screenly logout --name <profile>` removes a specific profile. Omitting `--name` removes
  the active profile. Removing the active profile leaves no profile active rather than
  promoting another one, so the next command can't silently talk to a different workspace.
- `screenly auth list` shows all stored profiles in a table with email and workspace name
  fetched from the API, with `*` marking the active profile. Honors `--output`.
- `screenly auth switch <name>` changes the active profile. Called without an argument,
  it lists the stored profile names and exits non-zero as a usage error.
- `screenly me` shows the email and workspace for the currently active token. Honors
  `--output`. Returns a clear error if not logged in or if the token is invalid.
- `API_TOKEN` still overrides the stored profiles. An empty value counts as unset.
- Plain text `~/.screenly` files (original format) are transparently migrated on first write.
  A file that is malformed in any other way is reported, not reinterpreted, and left in place.

## Storage

`~/.screenly` is written by creating a `0600` temp file and renaming it over the target, so
a reader never sees a torn store and a failed write leaves no temp file holding tokens
behind. Tokens live in a `BTreeMap`, so the file keeps a stable key order across rewrites.
The permissions are Unix-only and the write path takes no lock; both limits are documented
in the README.


sergey-borovkov (commented):
# Review — Multi-profile authentication support

## Overview

Replaces the plain-text `~/.screenly` single-token file with a YAML profile store, adding `login --name`, `logout --name`, `auth list`, `auth switch`, and `me` commands. Plain-text legacy files are transparently migrated. The migration design (auto-upgrade plain text → `default` profile on first write) is well thought out, and there's solid unit-test coverage for `Authentication` state transitions.

## Correctness

- **Token file permissions are not restricted.** The YAML file now holds *multiple* tokens but is still written via `fs::write` with default umask. The old behavior had the same flaw, but the blast radius is larger now (compromise of one file = all profiles). Consider `fs::set_permissions(&path, Permissions::from_mode(0o600))` on Unix after `write_store` (`src/authentication.rs` `write_store`).

- **`remove_token` post-delete active selection is non-deterministic.** When removing the active profile, `store.active = store.tokens.keys().next().cloned()` picks an arbitrary `HashMap` key — different on each run. Sort or pick deterministically (e.g., alphabetical first), and `info!()` the new active profile name so the user knows what just happened.

- **`Logout` panics on any error** via `.expect(\"Failed to remove token.\")` in `cli.rs`. Pre-existing pattern, but the new `remove_token` has more failure modes (NoCredentials, ProfileNotFound, Yaml parse). Better to print a friendly message and `exit(1)`, matching other auth paths.

- **`fetch_profile_info` assumes `/v4.1/users/me` returns an array.** `user.get(0).and_then(|u| u[\"email\"].as_str())` only works if the response is `[{...}]`, not `{...}`. The tests mock both shapes consistently, but please confirm the live API actually returns an array — if it returns an object, `me` will silently print `\"unknown\"`.

- **No User-Agent header on profile-info requests.** Existing `Authentication::client()` sets `screenly-cli {version}`; `fetch_profile_info` builds a bare client. Minor, but inconsistent.

- **Race on store updates.** `read → mutate → write` in `verify_and_store_token` / `remove_token` / `switch_profile` isn't atomic; two concurrent `screenly login` invocations can lose writes. CLI use makes this unlikely, but a temp-file + rename would be cheap insurance.

## Style / Conventions

- **Duplicated table-rendering block** in `AuthCommands::List` and `AuthCommands::Switch { name: None }` — ~25 identical lines. Extract a `print_profiles_table(profiles, api_url)` helper.

- **`AuthCommands::Switch` with no `name` printing the list is undocumented.** The clap help reads `<NAME> — Profile name to activate`, with no hint that omitting it is a meaningful action. Either document it (`/// Without an argument, prints the profile list.`) or make `name` required.

- **`Me` \"Profile:\" line shown conditionally** based on `active_profile_name()`. When `API_TOKEN` env var is in use, the field is silently omitted in both human and JSON output. Reasonable, but consider printing `Profile:   (from API_TOKEN env)` so the user understands why no profile name appears.

- **Redundant `use serde_yaml;`** at the top of `src/authentication.rs` — the `serde_yaml::Error` reference in the enum already pulls in the crate path.

- **`#[error(\"yaml error\")]`** swallows the underlying message. Use `#[error(\"yaml error: {0}\")]` so parse failures are diagnosable.

## Test coverage

Good additions for migration, switch, list, and `fetch_profile_info`. Missing:

- **`login` requires `--name` when multiple profiles exist** — the new safety check is the headline UX claim of the PR and isn't covered.
- **`verify_and_store_token` preserves existing profiles when called with a new `name`** — easy to regress if someone refactors to `TokenStore::default()`.
- **`remove_token(Some(name))` with explicit name** — only the `None` (active) path is tested.
- **Remove-active-profile picks a remaining one as new active** — would also catch the non-determinism above.

## Performance

- `auth list` and `auth switch` (no arg) make **2 sequential HTTP requests per profile** (`users/me` + `teams`). For 5+ profiles this is a noticeable hang. Consider parallelizing with `futures::join_all` (already a dep), or calling a combined endpoint if available.

## Security

- See file-permissions point above — primary concern.
- Tokens are passed around in `(name, token, is_active)` tuples returned from `list_profiles()`. Internal-only, but token strings are unused in the caller — consider returning `(name, is_active)` and exposing token lookup via a separate method to reduce accidental-print risk.

## Summary

Solid feature with a clean migration path. Top priorities before merge: deterministic active-profile selection on removal, file permissions (0o600), graceful `logout` error handling, and tests for the `--name` enforcement and explicit-name removal. The duplicate table code and undocumented `switch` no-arg behavior are easy follow-ups.

514sid:
Thanks @sergey-borovkov, good catches. Pushed a bunch of changes (and merged master in). Quick rundown:

Permissions and the write race are both handled now. `write_store` writes to a temp file and renames it into place, and chmods to 0600 on Unix. That kills two birds since the rename is atomic too.

Fixed the non-deterministic active selection on removal. It sorts the remaining profiles and takes the first one now, and `remove_token` returns the new active name so logout can actually tell the user what it switched to. Also killed the `.expect()` in logout, it prints a real message and exits 1 like the other auth paths.

On the users/me thing, I went and checked against prod with a real token. It does return an array, `[{...}]` with the email in it, and teams comes back as an array too, so the original indexing was fine. I made it fall back to reading an object anyway, just so we don't silently start printing "unknown" if that ever changes. Cheap insurance.

Rest of the smaller stuff: pulled the User-Agent into a shared client builder so `fetch_profile_info` isn't a bare client anymore, extracted the table rendering into one helper, documented the no-arg `switch`, and `me` now says `Profile: (from API_TOKEN env)` instead of just hiding the field. Dropped the stray `use serde_yaml` and the yaml error carries the underlying message now.

One thing I did differently from your suggestion: the client is blocking, so `futures::join_all` doesn't buy us anything. Used rayon instead (already a dep) to fan the per-profile requests out across threads. And `list_profiles` no longer hands tokens back to the caller, that all stays inside the auth module now.

Added tests for the `--name` enforcement, keeping existing profiles on a new login, removing by explicit name, and the deterministic active pick.

sergey-borovkov (commented):
## Review

Nice feature, and the storage layer is a real improvement over what it replaces: atomic tmp+rename write, `0600` permissions (the old file was `0644`), the trailing newline is now trimmed, and `logout` on a missing file no longer panics through `.expect()`.

I checked out the head commit to verify: `cargo clippy --all-targets -- -D warnings` is clean and all 18 `authentication::` tests pass. Everything below is from reading the code plus running the built binary against a fake `$HOME`.

### Blocking

1. **`screenly login` now fails for every existing user.** After migration everyone has exactly one profile, so plain `screenly login` — the normal flow after a token rotation — exits 1 with "A profile already exists". Details inline.
2. **A malformed store is silently reinterpreted as a raw token, and the profiles are destroyed on the next write.** Reproduced with a one-character typo. Details inline.

### Should fix

3. `me` ignores the global `--output` and reintroduces a second `--json` flag.
4. `me` reports the wrong profile name when `API_TOKEN` is set.
5. The profile table misaligns whenever a name or email is shorter than its header, or when info is missing.
6. `auth switch` with no argument swallows errors and exits 0.

### Minor

- Permission window on the tmp file, and a fixed tmp filename that two concurrent `login`s can race on (both inline).
- `NoCredentials` has no friendly message: `logout` no longer deletes the file, so the logged-out state now produces `NoCredentials` rather than `Io(NotFound)`, and the nice "Not logged in. Please run `screenly login`" branch is effectively dead for that path. Worth adding a `NoCredentials` arm.
- No per-profile API URL (inline) — worth deciding now while the format is new.
- `list_profiles` returning `Vec<(String, bool)>` reads poorly at call sites; a small struct (or reusing `ProfileEntry` with `info: None`) would be clearer.
- `Commands::Me` builds a `serde_json::Map` by hand; `serde_json::json!` is used elsewhere in the repo.

### Docs & tests

- `docs/CommandLineHelp.md` is regenerated correctly. `README.md` has no mention of profiles, `auth list`/`auth switch`, or `me`, which is where users will look first.
- Store-layer coverage is solid — migration, removal, deterministic re-activation, permissions, and both `users/me` response shapes. Gaps worth closing: `read_store` on malformed YAML (finding 2 — a test would have caught it), `fetch_profiles_with_info` (mockable with `$HOME` + `MockServer` like the existing tests), and `print_profiles_table` alignment (finding 5 — easy if it returns a `String` instead of `println!`-ing).
- Nothing asserts that a legacy plain-text file is actually rewritten as YAML on first write, despite that being a headline claim in the description.

src/cli.rs
```diff
@@ -549,13 +623,21 @@ pub fn handle_cli(cli: &Cli) {
     };
 
     match &cli.command {
-        Commands::Login {} => {
+        Commands::Login { name } => {
+            let existing = Authentication::list_profiles().unwrap_or_default();
+            let resolved_name = match resolve_login_name(name.as_deref(), &existing) {
+                Some(resolved) => resolved,
+                None => {
+                    error!("A profile already exists. Please specify a profile name with --name.");
```
sergey-borovkov: **Blocking: this breaks `screenly login` for every existing user.**

Verified against a legacy plain-text `~/.screenly`:

```
$ screenly login
ERROR A profile already exists. Please specify a profile name with --name.   (exit 1)
```

After migration every current user has exactly one profile, so the most common flow — re-logging in after a token rotation — is a hard break, as is any script or doc that says "run `screenly login`".

The guard protects against a case that isn't really ambiguous: re-logging in *should* update the profile you are on. Suggest `None => active_profile_name().unwrap_or("default")`, keeping the `saved under profile 'x'` message you already emit so the target is explicit.

Related: `Authentication::list_profiles().unwrap_or_default()` on the line above turns a store read error into "no profiles", which combined with the fallback below silently overwrites.

src/authentication.rs
```diff
@@ -57,6 +69,56 @@ impl Config {
...
+fn read_store() -> Result<TokenStore, AuthenticationError> {
+    let path = screenly_path()?;
+    if !path.exists() {
+        return Ok(TokenStore::default());
+    }
+    let contents = fs::read_to_string(&path)?;
+    if let Ok(store) = serde_yaml::from_str::<TokenStore>(&contents) {
+        return Ok(store);
+    }
+    // Backward compat: plain text token → migrate to "default" profile
```
sergey-borovkov: **Blocking: this fallback fires on *any* deserialization failure, not just genuine plain-text files — so a malformed store becomes a single bogus token and the real profiles are destroyed on the next write.**

Reproduced with a one-character typo (`tokenz:` instead of `tokens:`):

```
$ screenly auth list
  Profile  Email  Workspace
* default              <-- the real 'prod' profile is gone
$ screenly logout
$ cat ~/.screenly
active: null
tokens: {}             <-- prod_token_secret is now unrecoverable
```

`tokens` is a required field (a missing `active` is fine — serde defaults `Option` to `None`), so any hand-edit that touches it, any truncated file, or any future schema tightening silently converts credentials to garbage.

Gate the fallback on something that actually looks like a legacy token — a single non-empty line with no `:` and no newline — and return `AuthenticationError::Yaml` otherwise so the user can fix the file. That is also the one place the new `Yaml` variant would be useful; today it can only surface from `write_store`.

src/cli.rs
```diff
@@ -90,9 +134,26 @@ pub struct Cli {
...
-    Logout {},
+    Logout {
+        /// Profile name to remove. Removes the active profile if not specified.
+        #[arg(long)]
+        name: Option<String>,
+    },
+    /// Show information about the currently authenticated profile.
+    Me {
+        /// Enables JSON output.
+        #[arg(short, long, action = clap::ArgAction::SetTrue)]
```
sergey-borovkov: `master` recently standardized on the global `-o/--output table|json|csv`, with `--json` deprecated and hidden. A local `-j/--json` here means `screenly me --help` shows both flags and `--output json` is silently ignored:

```
Usage: screenly me [OPTIONS]
  -j, --json                Enables JSON output
  -o, --output <OUTPUT>     Output format: table (default), json, or csv
```

(clap skips propagating the global `json` because the subcommand shadows the id, so there is no panic — just two flags with different meanings.)

Suggest dropping the local flag and threading `output`/`OutputType` through, implementing `Formatter` for the profile info the way `Screens`/`Assets` do. The same applies to `auth list`, which is a table command that currently cannot emit JSON or CSV.

src/cli.rs
```diff
@@ -575,11 +657,97 @@ pub fn handle_cli(cli: &Cli) {
...
-        Commands::Logout {} => {
-            Authentication::remove_token().expect("Failed to remove token.");
-            info!("Logout successful.");
-            std::process::exit(0);
+        Commands::Me { json } => {
+            let auth = get_authentication();
+            match fetch_profile_info(&auth.token, &auth.config.url) {
+                Ok(info) => {
+                    let json_flag = json.unwrap_or(false);
+                    let profile = match active_profile_name() {
```
sergey-borovkov: This reports the wrong profile when `API_TOKEN` is set. `active_profile_name()` reads the store unconditionally, but `read_token()` returns the env token whenever `API_TOKEN` is present — so with both set, `me` prints `Profile: a` while actually authenticating as the env token, which is exactly backwards from the `(from API_TOKEN env)` label you fall back to. Check `env::var("API_TOKEN")` first.

src/cli.rs
```diff
@@ -28,12 +31,53 @@ fn get_authentication_error_message(e: &AuthenticationError) -> String {
...
+        None if existing.is_empty() => Some("default".to_string()),
+        None => None,
+    }
+}
+
+/// Prints the stored profiles as a table with email and workspace columns,
+/// marking the active profile with a `*`.
+fn print_profiles_table(entries: &[ProfileEntry]) {
+    let name_w = entries.iter().map(|e| e.name.len()).max().unwrap_or(0);
+    let email_w = entries
```
sergey-borovkov: The table misaligns whenever a name or email is shorter than its header, or when info is missing:

```
  Profile  Email  Workspace
  -    ---------
* a
  b
```

Two separate bugs:

- The widths ignore header length — use `max(len, "Profile".len())` and likewise for `Email`.
- `email_w` is `0` when every `info` is `None`, collapsing the column.

Rows with `info: None` also drop their columns entirely instead of printing a placeholder, and the user cannot tell whether the token is invalid or the network is down. Consider `-` or `(unavailable)` in both columns.

If this returned a `String` instead of `println!`-ing, it would be straightforward to unit-test.

src/cli.rs
```diff
@@ -575,11 +657,97 @@ pub fn handle_cli(cli: &Cli) {
...
+                }
+                Ok(entries) => print_profiles_table(&entries),
+                Err(e) => {
+                    error!("Error occurred: {e}");
+                    std::process::exit(1);
+                }
+            },
+            AuthCommands::Switch { name } => match name {
+                None => {
+                    if let Ok(entries) = fetch_profiles_with_info(&Config::default().url) {
```
sergey-borovkov: `if let Ok(entries) = ...` prints nothing and exits 0 when the fetch fails. It also does 2N network round-trips just to print a hint — `Authentication::list_profiles()` gives the same names offline.

Worth deciding deliberately too: `switch` with no argument silently behaving like `list` and exiting 0 is unusual for a CLI. A usage error with exit 1, plus the list, is friendlier to scripts.

src/authentication.rs
```diff
@@ -57,6 +69,56 @@ impl Config {
...
+}
+
+fn write_store(store: &TokenStore) -> Result<(), AuthenticationError> {
+    let path = screenly_path()?;
+    let contents = serde_yaml::to_string(store)?;
+
+    // Write to a temp file and rename over the target so a concurrent reader
+    // never observes a half-written store and a crash mid-write can't corrupt it.
+    let tmp_path = path.with_extension("tmp");
+    fs::write(&tmp_path, contents)?;
```
sergey-borovkov: `fs::write` creates the tmp file at `0644` and `restrict_permissions` chmods it afterwards, so the token is briefly world-readable. Creating it with `OpenOptions::new().mode(0o600)` closes the window.

Separately, `.screenly.tmp` (line above) is a fixed name shared by all processes, so two concurrent `login`s can interleave writes into it — which partly undercuts the atomicity the comment claims. A pid/random suffix fixes it, or `tempfile::NamedTempFile::persist`, since `tempfile` is already a dependency.

src/authentication.rs
```diff
@@ -29,10 +33,18 @@ pub enum AuthenticationError {
...
     #[error("invalid header error")]
     InvalidHeader(#[from] InvalidHeaderValue),
+    #[error("yaml error: {0}")]
+    Yaml(#[from] serde_yaml::Error),
     #[error("unknown error")]
     Unknown,
 }
 
+#[derive(Serialize, Deserialize, Default)]
+struct TokenStore {
```
sergey-borovkov, line 56: The store holds only tokens, so `auth list` sends *every* stored token to the single `API_BASE_URL`. Given that multi-profile mostly means multi-environment, a stage token gets transmitted to the prod host (and then shows blank because it 401s).

Storing a `url` per profile is worth considering now, while the format is new — it is much cheaper than a second migration later.

src/cli.rs
```diff
@@ -28,12 +31,53 @@ fn get_authentication_error_message(e: &AuthenticationError) -> String {
         AuthenticationError::Io(io_err) if io_err.kind() == std::io::ErrorKind::NotFound => {
             "Not logged in. Please run `screenly login` first to authenticate.".to_string()
         }
+        AuthenticationError::ProfileNotFound(name) => {
```
sergey-borovkov, line 43: Since `remove_token` no longer deletes the file, the ordinary logged-out state now yields `NoCredentials` rather than `Io(NotFound)`, so the friendly "Not logged in. Please run `screenly login` first" branch above is effectively dead for that path and users get the generic `Authentication error: no credentials error` instead. Worth adding a `NoCredentials` arm here.

src/cli.rs
514sid: Fixed in df627f5. `resolve_login_name` no longer errors when a profile already exists. Plain `screenly login` now updates the active profile (via `active_profile_name()`), falling back to `default` only on a fresh install, which is exactly the re-login-after-rotation flow you described. I also dropped the `list_profiles().unwrap_or_default()` bit, so a store read error can't masquerade as "no profiles" and trigger an overwrite. Covered by `test_resolve_login_name_defaults_to_active_profile`.

src/authentication.rs
514sid: Fixed in 6297cf6, with the error message made friendly in a50ee24. `read_store` now only migrates when the file actually looks like a legacy token (a single non-empty line with no `:`), gated by `is_legacy_token`. Any other parse failure returns `AuthenticationError::CorruptStore` instead of being reinterpreted, so the stored profiles survive. The error names the file path and tells the user to fix it or delete it and run `screenly login`, and it makes clear the file is left untouched.

I reproduced your `tokenz:` case against the built binary: it now exits 1 with that message and the file stays intact. Tests: `test_read_store_malformed_yaml_returns_error_not_token` and `test_legacy_plain_text_is_rewritten_as_yaml_on_first_write`.

src/cli.rs
514sid: Fixed in df627f5. Dropped the local `--json` flag. `Me {}` now threads the global `--output` through a `Formatter` impl (`ProfileDetails`) that supports table, json, and csv. Did the same for `auth list` via a `ProfilesTable` formatter, so it can emit json/csv too instead of being table-only.

src/cli.rs
514sid: Fixed in df627f5. The `me` label now checks `env::var("API_TOKEN")` first and prints `Profile: (from API_TOKEN env)` when the env token is in use, matching `read_token`'s precedence. So it no longer names a stored profile while actually authenticating as the env token. Verified live with both a stored profile and `API_TOKEN` set.

src/cli.rs
514sid: Fixed in df627f5. `print_profiles_table` became `format_profiles_table` and returns a `String`, so it is now unit-tested (`test_format_profiles_table_aligns_headers_and_placeholders`). Both column widths use `max(value_len, header_len)`, and a profile whose info could not be fetched renders `(unavailable)` in the email and workspace columns instead of dropping them. Verified the alignment live.

src/cli.rs
514sid: Fixed in df627f5. `auth switch` with no argument is now a usage error: it prints the available profile names and exits 1. The names come from `Authentication::list_profiles()` (the local store), so it does no network round-trips.

src/authentication.rs
514sid: Fixed in 6297cf6. `write_store` now creates the temp file through `create_private_file`, which opens it with mode `0o600` from the start, so there is no window where the token is world-readable. The temp filename also carries the process id, so two concurrent `login`s can't interleave into a shared `.screenly.tmp`. I didn't reach for `tempfile` here since it's only a dev-dependency in this crate.

src/authentication.rs
514sid: Good point, and I agree it's worth deciding while the format is new. I've deliberately left it out of this PR rather than rush it: doing it properly changes the store schema, the migration, `read_token`/`Authentication::new`, and `fetch_profiles_with_info`, and it hinges on a decision I'd rather make explicitly than guess at, namely how a profile's URL gets captured (a `login --url` flag, capturing `API_BASE_URL` at login time, or something else). I'd like to do it as its own follow-up PR with that decision settled. I can open a tracking issue if you want.

src/cli.rs
514sid: Fixed in df627f5. `get_authentication_error_message` now has a `NoCredentials` arm that returns the "Not logged in. Please run `screenly login` first to authenticate." message. Since logout leaves an empty store rather than deleting the file, the logged-out state surfaces as `NoCredentials`, and it now gets that friendly text instead of the generic error.

src/authentication.rs
sergey-borovkov: Confirmed fixed — I rebuilt at `a50ee24` and reproduced the `tokenz:` case: it exits 1 on both `auth list` and `logout`, and the file comes back byte-identical. `is_legacy_token` still migrates a genuine single-line token, and the error text is clear about what to do. Nice work on the rest of the thread too; I verified each of the other fixes against the built binary and they all behave as described.

One small residual gap in this change: a **zero-byte** `~/.screenly` now takes the `CorruptStore` path rather than being treated as an empty store.

```
$ : > ~/.screenly && screenly login
ERROR The credentials file at ~/.screenly could not be parsed (missing field `tokens`).
      Fix its contents, or delete it ... and run `screenly login` to start fresh.   (exit 1)
```

`is_legacy_token("")` is false, so an empty file falls through to the error. Two small annoyances: the message says "missing field `tokens`" for a file that has no fields at all, and it blocks `login` — the very command the message tells you to run — so the user has to delete the file first.

`read_store` already returns `TokenStore::default()` for a file that doesn't exist; a file that exists but is empty or whitespace-only should behave the same. Roughly:

```rust
if contents.trim().is_empty() {
    return Ok(TokenStore::default());
}
```

Non-blocking from my side — it's recoverable since the error names the file, so fix it or wave it through as you prefer.

514sid: Good catch, fixed in 855bca3. `read_store` now returns an empty store for a zero-byte or whitespace-only file, the same as a missing one, so it no longer takes the `CorruptStore` path and no longer blocks `login`. Verified live: `me` on an empty file now reports "Not logged in" and `login` proceeds. Added a test covering both empty and whitespace-only contents.

Thanks for re-verifying the rest against the binary.

src/authentication.rs
514sid: Opened #303 to track this as a follow-up, including the open decision on how a profile's URL should be captured (`--url` flag vs. capturing the resolved base URL at login time). Keeping it out of this PR so we can settle that separately.

514sid:
@sergey-borovkov thanks for the thorough re-review. Everything from this round is in:

- Empty/whitespace-only `~/.screenly` is now treated as an empty store instead of a corrupt one, so it no longer blocks `login` (855bca3, with a test for both empty and whitespace-only files).
- The per-profile API URL is tracked separately in #303, including the open decision on how a profile's URL gets captured.

All the earlier fixes (login default, malformed-store guard, `me`/`auth list` output, table alignment, `auth switch` no-arg, tmp file hardening, NoCredentials message) are pushed too. 210 tests pass, clippy and fmt clean, and I re-ran the full auth matrix (login/logout/me/auth list/switch, migration, corrupt and empty stores, API_TOKEN override, output formats) against the built binary.

Ready for another look when you have a moment.

PR #300, opened 2026-07-20
Harden CI supply chain and add grouped Dependabot config

vpetersson:
CI supply-chain hardening, plus the dependency-update plumbing that prompted it.

## Background

Five Dependabot PRs (#280, #283, #288, #289, #290) sat open for up to three months. All five turned out to be already superseded by #297 (`7531baa`) and #299 (`5dad863`) — every target version was at or below what `Cargo.lock` already had. They've been closed. This PR addresses why they accumulated, and what turned up while looking.

### 1. Add `.github/dependabot.yml`

The repo had no Dependabot config at all, so security updates ran with default settings and opened one PR per advisory.

- **Cargo — security updates grouped** into a single PR. Groups default to version updates, so `applies-to: security-updates` is required.
- **Cargo — version updates stay disabled** via `open-pull-requests-limit: 0`. This repo takes CVE-driven bumps only; enabling routine version updates would *add* PR volume rather than reduce it. Deliberate no-change to current behavior.
- **GitHub Actions — grouped**, both version and security. Not previously covered at all.

### 2. Drop unmaintained / third-party actions

- `actions-rs/clippy-check@v1.0.7` → `cargo clippy` directly. The actions-rs org has been archived since 2021 and the action runs on the deprecated Node 12 runtime.
- `dtolnay/rust-toolchain` → `rustup`, in `docs.yml`, `release.yml`, and `fmt.yml`. The runners have rustup preinstalled. **No `dtolnay` references remain.**

Permissions on the lint job drop to `contents: read` — `pull-requests: write` and `checks: write` existed only so the action could post check annotations.

### 3. Pin all third-party actions to commit SHAs

Nothing was digest-pinned. Four actions ran from **mutable branch refs** — `dtolnay/rust-toolchain@master`, `@nightly`, `sbomify/github-action@master`, `DeterminateSystems/flakehub-cache-action@main` — meaning any upstream push executed automatically. Two of those run in `release.yml` and `sbom.yml`, which build and attest release artifacts.

Every third-party reference is now pinned to a full SHA with the version as a trailing comment. Dependabot updates digest pins and maintains those comments, which is what the grouping in (1) is for.

`screenly/cli@master` is intentionally left unpinned: `actions.yml` exercises the CLI at master by design.

### 4. Upgrade outdated actions

- `actions/checkout` v3 → v4 in `docs.yml` and `release.yml`, matching the other workflows. v3 runs on deprecated Node 16.
- `softprops/action-gh-release` v1 → v3.0.2. Also a Node 16 action; v3 runs Node 24. Input-compatible — every input this workflow uses (`files`, plus the `GITHUB_TOKEN` env) still exists in v3, with only additions.
- `sbomify/github-action` `master` → `v26.7.0`. `action.yml` is byte-identical between the two, so the interface is unchanged. Note this moves *back* 10 commits — from an untagged branch tip onto the latest release, which is the point.
- `flakehub-cache-action` needed no change: its pinned SHA is already identical to `v3.21.7`, so only the trailing comment was corrected.

### 5. Fix 19 clippy warnings and enforce the gate

`useless_borrows_in_formatting` has fired since clippy 1.97 but went unnoticed, because the old actions-rs step only annotated warnings rather than failing on them. Removed 19 redundant `&` in `format!`/`debug!`/`println!` args across six files via `cargo clippy --fix` — 19 insertions, 19 deletions, inert since formatting traits auto-deref.

With the tree clean, the clippy step now runs `-D warnings` so these can't silently accumulate again.

### 6. Path-filter fix

`lint.yml` was added to its own `paths` filter. Previously, edits to the lint workflow didn't run it, so a broken lint config could merge unnoticed.

## Verification

All 9 checks pass. Locally, against clippy/rustc **1.97.1** (matching the 1.97 the runner ships):

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test` — 200 passed, 0 failed
- `cargo fmt --all --check` — clean
- `cargo build --all-targets` — clean
- All workflow YAML parses

## Reviewer notes — please read before merging

**`release.yml` and `sbom.yml` trigger only on `v*` tags, so no PR check exercises them.** Green CI on this PR says nothing about either. This branch changes the release pipeline in three unverified ways:

1. `dtolnay/rust-toolchain` → `rustup` across the 8-target build matrix
2. `actions/checkout` v3 → v4
3. `action-gh-release` v1 → v3.0.2 (two major versions)

A throwaway pre-release tag is the only way to confirm these before a real release depends on them. **This is the main risk in the PR.**

**Pinning `sbomify/github-action` is largely cosmetic.** Its `action.yml` runs `docker://ghcr.io/sbomify/sbomify-action:latest` — a mutable Docker tag. Whatever ref is pinned here, the executed code is whatever `:latest` resolves to at run time. `sbom.yml` therefore still has an unpinned supply chain, and it can't be fixed from this side; mitigating it means forking or replacing the action. This is the one real gap remaining.

**`rustfmt.toml`'s unstable options are now inert.** `group_imports` and `imports_granularity` only applied under nightly; `fmt.yml` now uses stable. Ordinary formatting drift is still caught, but import ordering is no longer enforced. Either delete them as dead config or keep them for developers who run nightly by choice.

**`-D warnings` with a floating toolchain will break again** when a future clippy adds lints — that's exactly how the 19 warnings here surfaced. It's the tradeoff for the gate being meaningful; the alternative is pinning the clippy toolchain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,34 @@
+version: 2
+updates:
+  # Cargo: security updates only, bundled into a single PR.
+  #
+  # This repo does not take routine version bumps -- `open-pull-requests-limit: 0`
+  # disables version updates while leaving Dependabot security updates enabled.
+  # Without the group below, each advisory opens its own PR (see #280, #283,
+  # #288, #289, #290, all of which sat open for months after the underlying
+  # bumps had already landed).
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 0
+    groups:
+      cargo-security:
+        applies-to: security-updates
+        patterns:
+          - "*"
+
+  # GitHub Actions: group both version and security updates into one PR each.
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
+    groups:
+      actions-version:
+        applies-to: version-updates
+        patterns:
+          - "*"
+      actions-security:
+        applies-to: security-updates
+        patterns:
+          - "*"
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     name: List screens
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - id: list-screens
         uses: screenly/cli@master
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,10 +43,10 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master
-        with:
-          toolchain: stable
-          target: x86_64-unknown-linux-gnu
+        run: |
+          rustup toolchain install stable --profile minimal
+          rustup default stable
+          rustup target add x86_64-unknown-linux-gnu
 
       - name: Generate documentation
         run: cargo run -- print-help-markdown > /tmp/CommandLineHelp.md
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
       - name: Install Rust
         run: |
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,8 +20,8 @@ jobs:
       docs_command_line_help: ${{ steps.filter.outputs.docs_command_line_help }}
       docs_edge_apps: ${{ steps.filter.outputs.docs_edge_apps }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3
         id: filter
         with:
           list-files: shell
@@ -40,10 +40,10 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master
         with:
           toolchain: stable
           target: x86_64-unknown-linux-gnu
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out Git repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           submodules: recursive
           ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: dtolnay/rust-toolchain@4fd1da8b0805d2d2e936788875a7d65dbd677dc2 # nightly
         with:
           components: rustfmt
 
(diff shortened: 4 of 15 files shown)


sergey-borovkov (commented):
**Verification:** all 9 pinned SHAs resolve to their claimed tags (including `softprops/action-gh-release` v3.0.2, whose annotated tag derefs to `3d0d988`). No wrong or unknown digests. The Rust changes are pure no-ops — `&` before a formatting arg is inert since format traits auto-deref.

Solid, well-scoped hardening pass, and the description is unusually honest about its own gaps. Most findings below are things the pass stopped just short of.

---

### 1. `fmt.yml`: the regression is bigger than the description says

The description flags that `group_imports`/`imports_granularity` go inert. The CI log from this PR shows **all five** `rustfmt.toml` options are dropped:

```
Warning: can't set `imports_granularity = Module` ...
Warning: can't set `group_imports = StdExternalCrate` ...
Warning: can't set `format_generated_files = false` ...
Warning: can't set `unstable_features = true` ...
Warning: can't set `ignore = IgnoreList { path_set: {"furiousfilter/**"} } ...
```

So in CI, `rustfmt.toml` is now effectively an empty file. That's a semantic change to a merge gate riding along in a supply-chain PR — and it isn't required to reach the PR's goal, since dropping `dtolnay` doesn't require dropping nightly:

```yaml
- name: Install nightly rustfmt
  run: rustup toolchain install nightly --profile minimal --component rustfmt
- name: Run rustfmt check
  run: cargo +nightly fmt --all --manifest-path Cargo.toml --check
```

This removes the third-party action *and* preserves behavior exactly. I'd prefer it over the "delete the dead config" option — deleting it is a real policy change that deserves its own PR rather than being a side effect of pinning actions.

### 2. `actions/attest-build-provenance` left three majors behind

Pinned to `ef24412` = **v1.4.4**, commented `# v1`; latest is **v4.1.1**. This is the action producing provenance attestations for release artifacts and the SBOM — the one place in the repo where staleness has security meaning. The PR upgrades `checkout` and `action-gh-release` on exactly this rationale but skips this one.

Related: this PR's own CI log emits

```
Node.js 20 is deprecated ... actions/checkout@34e114876b... forced to run on Node.js 24
```

`checkout` v4.3.1 targets Node 20, now deprecated — the same argument used for v3→v4 applies to v4→v5. Either go to v5, or state explicitly that v4 is the intended floor.

### 3. `rust.yml` runs `rust:latest` — a fixable instance of the gap you called out

`build_and_test` runs `docker run ... rust:${{ matrix.rust }}` with `matrix.rust: [latest, 1.88.0]`. That's a mutable Docker tag executing against PR-authored code on every push — structurally the same problem as `sbomify`'s `docker://...:latest`, except this one *is* fixable from this repo (pin a digest, or name a concrete version). Relatedly, `dependabot.yml` has no `docker` ecosystem entry despite the repo shipping a Dockerfile.

### 4. The path-filter fix was applied to `lint.yml` only

`fmt.yml` has the identical bug — its `paths` filter is `**/*.rs`, so edits to `fmt.yml` don't run it. Same one-line fix.

`fmt.yml` also still carries `permissions: pull-requests: write`, which — like the lint permissions this PR correctly drops — nothing in the job uses.

Minor: `lint.yml`'s `paths` covers `**.rs` and `Cargo.toml` but not `Cargo.lock`, so a lockfile-only dependency bump — i.e. exactly what the new grouped Dependabot PRs will be — skips clippy entirely.

### 5. Nits

- **Version-comment precision is inconsistent.** `# v4` is really v4.3.1, `# v3` is v3.0.3, `# v1` is v1.4.4 — while `# v3.0.2`, `# v26.7.0`, `# v3.21.7` are exact. The comment is the only human-readable check on a digest, and Dependabot rewrites these to exact versions on update anyway, so I'd make them exact now rather than let the style drift.
- **Dead config worth a follow-up:** there's no `.gitmodules` and no `furiousfilter` path in this repo, so `rustfmt.toml`'s `ignore` is already vestigial — as are `submodules: recursive` and `ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}` in the `fmt.yml`/`lint.yml` checkouts. That last one hands a private SSH key to two jobs that don't need it, which is odd company for a hardening PR.
- **Pre-existing, but adjacent to touched lines:** `debug!("Response: {:?}", response.text()?)` in `src/commands/mod.rs` (`post`, `delete`, `patch`) puts a `?` inside a `log` macro. `debug!` only evaluates its args when the level is enabled, so whether that error propagates depends on the log level.

---

### `dependabot.yml` — correct as written

`open-pull-requests-limit: 0` does disable version updates without affecting security updates, and `applies-to: security-updates` is genuinely required for the group to apply to them. No issues.

### On the release-pipeline risk

Your reviewer note is right. The `rustup` replacement looks faithful to what `dtolnay/rust-toolchain` did (install with `--profile minimal`, set default, add target), and rustup is preinstalled on both `ubuntu-22.04` and `macos-15` — but "looks faithful" is doing real work across an 8-target matrix plus two unverified action major bumps. A throwaway `v*-rc` tag is cheap insurance.

**Recommendation:** address (1) before merge — three lines, and it's a gate weakening. (2), (3), (4) are worth folding in now since the PR is already in those files. The rest can be follow-ups.

sergey-borovkov (changes requested):
Blocking on finding (1) in my review above: `fmt.yml`'s switch to stable rustfmt silently disables all five `rustfmt.toml` options (confirmed in this PR's own CI log), which weakens a merge gate as a side effect of a supply-chain PR. The three-line `cargo +nightly fmt` fix keeps the `dtolnay` removal while preserving behavior exactly.

Everything else in the review is non-blocking.

PR #311, opened 2026-08-25
docs: document depends_on conditional visibility for settings

nicomiguelino:
## Summary
- Documents the new `properties.depends_on` key in a setting's `help_text` schema, which the web console (Screenly/Screenly#2554) now supports for conditional visibility: a setting only renders (and is submitted) while another named setting's current value is one of a listed set of values.
- Adds a "Conditional visibility" example alongside the existing input field type examples.
- No Rust changes needed: `help_text` stays an opaque string to the CLI, exactly like `type`/`options`/`advanced` already are.

### DIFF
--- a/docs/EdgeApps.md
+++ b/docs/EdgeApps.md
@@ -483,7 +483,7 @@ Edge App settings support additional input field types beyond plain text and pas
   - `properties.type`: One of `datetime`, `number`, `select`, `boolean`, `textarea`, `url`.
   - `properties.help_text`: Human-friendly description shown in the UI.
   - `properties.options` (only for `select`): Array of `{ label, value }` options.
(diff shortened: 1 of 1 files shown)


sergey-borovkov (changes requested):
Docs read well and the example is clear. One gap I'd like closed before merging.

### `depends_on` on a required field is undefined

`docs/EdgeApps.md:486`

The bullet says a hidden field's value "isn't collected or written on save", and the example correctly uses `optional: true` — but nothing states whether `optional: false` combined with `depends_on` is a supported combination.

A reader who marks a conditional field required gets an install form that may fail validation on a field they cannot see, with no way to work out why from these docs. One sentence either saying conditional fields should be `optional: true`, or describing how required-ness is skipped while the field is hidden, closes it.

---

Note for merge ordering: this PR and #308 both add a bullet to the same list at `docs/EdgeApps.md:486` and a section at the same anchor, so expect a textual conflict on whichever lands second. No semantic conflict — #308's `help_text_with_display_order` only inserts into an existing schema's `properties`, so it preserves `depends_on`.

sergey-borovkov (approved):
The added sentence covers it — thanks. Approving.

Heads up that this will still conflict textually with #308 at `docs/EdgeApps.md:486` (same list, same anchor); whichever lands second needs a trivial rebase. No semantic conflict.

PR #317, opened 2026-09-02
Document validation key for Edge App settings

nicomiguelino:
## Summary
- Documents `properties.validation` as a new structured `help_text` key for Edge App settings: an optional regular expression the dashboard enforces against the setting's value, on blur and on save.
- Adds a "Pattern validation" example alongside the existing input field type examples.
- No Rust changes needed: `help_text` stays an opaque string to the CLI, exactly like `type`/`options`/`display_order`/`depends_on` already are.

## Context
- `validation` piggybacks on the existing structured `help_text` schema rather than becoming a new column, so no backend schema/migration is required for this to work.
- Deploy-time linting (pattern length cap, rejecting backreferences/nested unbounded quantifiers, rejecting `validation` on `select`/`boolean` settings) and the dashboard-side rendering/enforcement are intentionally out of scope for this PR, this documents the manifest-authoring contract only.

## Test plan
- [x] `cargo test` — all tests pass except one pre-existing, unrelated failure (`authentication::tests::test_read_token_correct_token_is_returned`), which reproduces identically on `master`.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy --bin screenly --all-targets -- -D warnings` clean.


docs/EdgeApps.md
```diff
@@ -487,6 +487,7 @@ Edge App settings support additional input field types beyond plain text and pas
...
   - `properties.depends_on`: Optional `{ setting, values }` object that makes this field's visibility depend on another setting's current value. The field only renders (and is submitted) while `setting`'s current value is one of `values`, otherwise it's hidden and skipped. A malformed or stale reference (a typo in `setting`, or a setting later renamed or removed) fails open, so the field stays visible rather than disappearing. A field with `depends_on` can still be marked `optional: false`; its required-ness is only enforced while the field is visible, and is skipped along with the rest of validation while it's hidden.
+  - `properties.validation`: Optional regular expression the value must match. The pattern is implicitly anchored to a full match (`[A-Z]{3}` means the whole value is three uppercase letters, not that it contains them somewhere). An empty value on an optional field skips this check; on a required field the required check takes precedence. This is enforced by the dashboard only, on blur and on save; it is not enforced by the CLI or the API, so an Edge App must still treat setting values as untrusted input.
```
sergey-borovkov: The "implicitly anchored to a full match" claim is a contract with the dashboard that nothing in this repo enforces or tests, and the example below depends on it being true. If the dashboard implements this as `new RegExp(pattern).test(value)` rather than an HTML `pattern` attribute, the check is unanchored and `XLHRXX` passes an "airport code" field - silently under-validating for every author who copies the example.

Writing the examples explicitly anchored (`'^[A-Z]{3}$'`) is correct under *both* semantics - an HTML `pattern` wraps it as `^(?:^[A-Z]{3}$)$`, which still matches - so it costs nothing and removes the dependency on an unverified claim.

docs/EdgeApps.md
```diff
@@ -592,6 +593,23 @@ settings:
...
+    title: Airport Code
+    optional: false
+    help_text:
+      schema_version: 1
+      properties:
+        help_text: "IATA code for the departure board. Three uppercase letters, like LHR."
+        validation: '[A-Z]{3}'
+```
+
+`validation` is not supported on `select` or `boolean` fields, since the widget itself already constrains the value.
```
sergey-borovkov: "not supported" reads like it will be rejected, but nothing rejects it. There's no deploy-time lint (out of scope per the PR description), so a manifest with `validation` on a `select` deploys cleanly and the key is just ignored by the dashboard. Suggest saying it is *silently ignored*, so authors don't spend time debugging a rule that never fires.

This is also the only place the restriction appears - the schema bullet list up at line 490 doesn't carry it, so someone reading only the key reference won't see it.

src/commands/edge_app/manifest.rs
```diff
@@ -586,6 +586,53 @@ settings:
         assert_eq!(actual, expected);
     }
 
+    #[test]
+    fn test_manifest_allows_validation_in_structured_help_text() {
```
sergey-borovkov: This test can't fail for a `validation`-specific reason. `deserialize_help_text` accepts any JSON object wholesale, `serialize_help_text` re-emits any object wholesale, and `help_text_with_display_order` only ever touches `display_order` and `schema_version` - no code path anywhere knows the string `"validation"`.

It's `test_manifest_allows_structured_help_text` (parse assertion) plus `test_save_manifest_to_file_serializes_structured_help_text` (save assertion) with the payload swapped, so it only fails if generic structured-`help_text` handling breaks - which those two already cover. Suggest dropping it and adding a real test when the CLI actually starts inspecting `validation` (i.e. when the deferred deploy-time lint lands).

sergey-borovkov (commented):
Reviewed the docs against the code paths (`serialize_help_text`/`deserialize_help_text`/`help_text_with_display_order` in `src/api/edge_app/setting.rs`, `new`/`save_to_file` in `src/commands/edge_app/manifest.rs`, `format_yaml` in `src/commands/serde_utils.rs`).

The core claim checks out: `help_text` is opaque to the CLI, so `validation` survives parse -> `assign_setting_display_orders` -> `save_to_file` untouched. I also round-tripped seven regexes containing backslashes, `#`, `'`, `[`, `{` and tabs through a scratch test - all byte-identical. No correctness bug here.

Three non-blocking notes inline.

docs/EdgeApps.md
```diff
@@ -491,6 +491,7 @@ Edge App settings support additional input field types beyond plain text and pas
...
+  - `properties.validation`: Optional regular expression the value must match. The pattern is implicitly anchored to a full match (`[A-Z]{3}` means the whole value is three uppercase letters, not that it contains them somewhere); write it explicitly anchored (`^[A-Z]{3}$`) anyway, since that's correct regardless of how the anchoring is implemented. An empty value on an optional field skips this check; on a required field the required check takes precedence. Not supported on `select` or `boolean` fields, since the widget itself already constrains the value; there's no deploy-time check, so `validation` on those types is silently ignored rather than rejected. This is enforced by the dashboard only, on blur and on save; it is not enforced by the CLI or the API, so an Edge App must still treat setting values as untrusted input.
```
sergey-borovkov: The bullet is admirably honest that an unsupported `validation` is silently ignored on `select`/`boolean`, but it says nothing about a **syntactically invalid** regex. Neither the CLI (no code path reads this key) nor the API checks it, so a typo like `'^[A-Z{3}$'` deploys cleanly and only fails when the dashboard calls `new RegExp` on it — potentially breaking the settings form for that app, with no signal at deploy time. One sentence saying malformed patterns are not caught anywhere would match the honesty of the rest of the bullet.

sergey-borovkov: "implicitly anchored to a full match" is a claim about dashboard behavior that nothing in this repo can pin down, and the same sentence then hedges with "regardless of how the anchoring is implemented" — which reads as hedging against an unverified claim. Suggest dropping the implicit-anchoring sentence and simply requiring explicit anchors. As written, an author who trusts that line and omits `^`/`$` gets substring matching if the dashboard ever uses `RegExp.test` unwrapped.

docs/EdgeApps.md
```diff
@@ -596,6 +597,21 @@ settings:
...
+```yaml
+settings:
+  airport_code:
+    type: string
+    title: Airport Code
+    optional: false
+    help_text:
+      schema_version: 1
+      properties:
+        help_text: "IATA code for the departure board. Three uppercase letters, like LHR."
```
sergey-borovkov, line 612: This is the only structured `help_text` descriptor in the docs that omits `properties.type`, and the "Common keys" list above never says `type` is optional. A reader copying this example for a plain-text field can't tell whether the descriptor is valid without `type`, or whether they need to add some `string`-ish value (which isn't in the documented enum). Either add a `type` here or add one clause upstream saying `type` may be omitted for a plain text input.

sergey-borovkov (dismissed):
Approving — docs-only, and I verified the load-bearing CLI claims against the code:

- `help_text` really is opaque to the CLI: `deserialize_help_text` / `serialize_help_text` in `src/api/edge_app/setting.rs` accept either a plain string or an arbitrary JSON object and round-trip it verbatim. `deny_unknown_fields` is on `Setting`, not on the `properties` map, so an unknown `validation` key is neither rejected nor dropped.
- `help_text_with_display_order` / `assign_setting_display_orders` only insert `display_order` and `schema_version` into an existing `properties` object and preserve sibling keys, so `validation` survives deploy intact.
- Nothing in the repo enumerates the allowed `properties.*` keys, so "no deploy-time check, silently ignored on `select`/`boolean`" is accurate.
- The new YAML example matches the shape already covered by `test_manifest_allows_structured_help_text` and `test_save_manifest_to_file_serializes_structured_help_text`, so it parses and re-serializes correctly.

Three non-blocking documentation-accuracy notes inline. One housekeeping item: the PR description still says "a new test proves a manifest with `properties.validation` round-trips correctly" — that test was added in 2bf48aa and removed again in fe10296, so the net diff is docs-only and the description should be updated before merge.

docs/EdgeApps.md
nicomiguelino: Upon checking the task requirements, documenting the pattern as implicitly anchored to a full match is intentional and part of the requirements for this feature. We're keeping that wording and not adding a "write it explicitly anchored" instruction alongside it, since the anchoring behavior itself is the documented guarantee.

docs/EdgeApps.md
nicomiguelino: I agree on this one. I'll add the `type` properly in future commit(s).

sergey-borovkov (approved):
Re-approving on 13c47b3. `properties.type` now documents `string` as the omittable default and the example carries it, and the validation bullet now states that malformed patterns deploy cleanly and surface only in the dashboard — both earlier comments addressed.

On anchoring you went the other way, keeping the implicit-anchoring claim and de-anchoring the example to `'[A-Z]{3}'`. Nothing in this repo can confirm that behavior (no `properties.*` value is read anywhere in the CLI), so it rests entirely on the dashboard. If you have confirmed it there, fine by me — flagging only so the choice is on the record.

Reminder: the PR description still advertises a round-trip test that was removed in fe10296; the net diff is docs-only.

docs/EdgeApps.md
```diff
@@ -486,11 +486,12 @@ Edge App settings support additional input field types beyond plain text and pas
...
   - `properties.depends_on`: Optional `{ setting, values }` object that makes this field's visibility depend on another setting's current value. The field only renders (and is submitted) while `setting`'s current value is one of `values`, otherwise it's hidden and skipped. A malformed or stale reference (a typo in `setting`, or a setting later renamed or removed) fails open, so the field stays visible rather than disappearing. A field with `depends_on` can still be marked `optional: false`; its required-ness is only enforced while the field is visible, and is skipped along with the rest of validation while it's hidden.
+  - `properties.validation`: Optional regex the value must match, checked by the dashboard on blur and on save. Patterns are implicitly anchored to a full match (`[A-Z]{3}` matches only exactly three uppercase letters). Empty values skip this check when the field is optional; otherwise the required check takes precedence. Silently ignored on `select`/`boolean` fields, and never checked by the CLI or API, so unsupported types and malformed patterns alike deploy cleanly and surface only in the dashboard. Treat setting values as untrusted input regardless.
```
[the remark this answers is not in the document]
nicomiguelino: Updated the PR description to drop that stale claim, the test was removed earlier per Sergey's review and never restored in the description. This PR is docs-only now, no code paths reference `validation`.

PR #320, opened 2026-09-06
Fix help text column showing raw JSON for structured settings

nicomiguelino:
## Summary

`screenly edge-app setting list` printed the raw JSON of `help_text` for settings that use the structured help text format, instead of the human readable text. This extracts `properties.help_text` from the structured format before rendering the table.

## Before

```
+------------------------+------------------------+-------+---------------+----------+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Name                   | Title                  | Value | Default value | Optional | Type   | Help text                                                                                                                                                                                                                                                                                                      |
+------------------------+------------------------+-------+---------------+----------+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| enable_analytics       | Enable Analytics       |       | true          | Yes      | string | {"properties":{"display_order":0,"help_text":"Enable or disable Sentry and Google Analytics integrations."},"schema_version":1}                                                                                                                                                                                |
+------------------------+------------------------+-------+---------------+----------+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| style                  | Style                  |       | traditional   | Yes      | string | {"properties":{"display_order":4,"help_text":"Select the visual style for the clock.","options":[{"label":"Traditional","value":"traditional"},{"label":"Modern","value":"modern"},{"label":"Minimal","value":"minimal"}],"type":"select"},"schema_version":1}                                                 |
+------------------------+------------------------+-------+---------------+----------+--------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

## After

```
+------------------------+------------------------+-------+---------------+----------+--------+---------------------------------------------------------------------+
| Name                   | Title                  | Value | Default value | Optional | Type   | Help text                                                            |
+------------------------+------------------------+-------+---------------+----------+--------+---------------------------------------------------------------------+
| enable_analytics       | Enable Analytics       |       | true          | Yes      | string | Enable or disable Sentry and Google Analytics integrations.          |
+------------------------+------------------------+-------+---------------+----------+--------+---------------------------------------------------------------------+
| style                  | Style                  |       | traditional   | Yes      | string | Select the visual style for the clock.                               |
+------------------------+------------------------+-------+---------------+----------+--------+---------------------------------------------------------------------+
```

Plain string `help_text` values (e.g. `override_locale`, `override_timezone`) are unaffected.

## Edge case

A structured `help_text` with no `properties.help_text` key now renders an empty cell instead of falling back to the raw JSON.


src/api/edge_app/setting.rs
```diff
@@ -270,6 +270,18 @@ fn is_structured_help_text(help_text: &str) -> bool {
...
 }
 
+pub(crate) fn extract_display_help_text(help_text: &str) -> String {
+    match serde_json::from_str::<Value>(help_text) {
+        Ok(Value::Object(object)) => object
+            .get("properties")
+            .and_then(|properties| properties.get("help_text"))
+            .and_then(|value| value.as_str())
+            .map(|value| value.to_string())
+            .unwrap_or_else(|| help_text.to_string()),
```
sergey-borovkov: This fallback means a structured help_text with **no** `properties.help_text` still prints the raw JSON blob — the exact output this PR set out to remove.

Nothing requires that key. A setting declared as:

```yaml
help_text:
  schema_version: 1
  properties:
    type: number
```

passes `deserialize_help_text` and `has_malformed_properties`, gets a `display_order` injected by `help_text_with_display_order`, and then lands here — rendering `{"properties":{"display_order":0,"type":"number"},"schema_version":1}` in the Help text column.

That isn't hypothetical: `properties.type`, `properties.options`, and `properties.validation` are all documented as independent keys, so a `select` with options and no description is an ordinary manifest. Suggest rendering an empty cell when the value parses as an object but carries no `properties.help_text` string, and reserving the verbatim fallback for the non-object case that already falls to the `_` arm below.

src/commands/mod.rs
```diff
@@ -429,6 +429,12 @@ impl Formatter for EdgeAppSettings {
                         }
                         return Cell::new("");
                     }
+                    if field_name.eq("help_text") {
+                        let help_text = field_value.as_str().unwrap_or_default();
```
sergey-borovkov: Minor / defensive, and explicitly **not** introduced by this PR — the unchanged fallback four lines down does the same `as_str().unwrap_or_default()`.

Extraction is keyed on `field_value.as_str()`, so it only handles help_text arriving as a *stringified* JSON blob. `EdgeAppSettings` holds the raw response `Value` and never passes through `deserialize_help_text`, whose `Structured(Value)` arm shows the object form is a shape this codebase does contemplate. If the settings endpoint ever returned `"help_text": {…}`, `as_str()` would be `None` and the column would silently render empty.

The PR's own Before/After output shows the string form, so this looks inert today. Cheap hardening if you ever touch it again: have the helper take `&serde_json::Value` and match `Value::Object` directly instead of `&str`.

sergey-borovkov (dismissed):
Approving — the fix is correct for the case it targets and I found no regression.

What I checked: `extract_display_help_text` only rewrites the cell when the value parses to a JSON object **and** `properties.help_text` is a string; every other input returns verbatim, so plain help texts and the `HELP_TEXT_NAME_OVERRIDES` settings are untouched. JSON output goes through the raw-value branch of `format_value` and is unaffected, and `EdgeAppSettings::supports_csv()` is `false` so the CSV branch is unreachable. Built the branch, ran the full suite (235 passed, including the new test) and `cargo clippy --all-targets` — clean.

Two non-blocking notes inline. The first is a real gap in this fix rather than a style point, and it's a couple of lines in the function this PR just added, so it may be worth folding in here rather than as a follow-up.

sergey-borovkov (dismissed):

salmanfarisvp (approved):
As per slack discussion.

PR #326, opened 2026-09-22
Add whoami command using GET /v3/me

salmanfarisvp:
## Summary
- Adds a top-level `screenly whoami` command that calls `GET /v3/me` in a single request
- Displays user email, name, IDs, and workspace details in table, JSON, and CSV output formats
- Updates CLI help docs and adds unit tests with non-personal fixtures

## Local build test 

- Values are redacted 

```
> screenly whoami   

+---------------+-----------------------------+
| Field         | Value                       |
+---------------+-----------------------------+
| Email         | name@screenly.io            |
+---------------+-----------------------------+
| Name          | Salman Faris                |
+---------------+-----------------------------+
| User ID       | xxxxxxxxxxxxxxxxxxxxxxxxxxx |
+---------------+-----------------------------+
| Workspace     | CSS                         |
+---------------+-----------------------------+
| Workspace ID  | xxxxxxxxxxxxxxxxxxxxxxxxxxx |
+---------------+-----------------------------+
| Workspace URL |https://salman.screenlyapp.com|
+---------------+-----------------------------+
```


src/commands/whoami.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn full_name(&self) -> String {
+        let first = self.field("user", "first_name");
+        let last = self.field("user", "last_name");
+        match (first.as_str(), last.as_str()) {
+            ("N/A", "N/A") => "N/A".to_string(),
+            ("N/A", last) => last.to_string(),
+            (first, "N/A") => first.to_string(),
+            (first, last) => format!("{first} {last}"),
+        }
+    }
```
sergey-borovkov: `"N/A"` is doing double duty here, as both the display text and the "missing" signal, which leads to a few edge cases:

- Empty strings aren't treated as missing. With `first_name: "", last_name: ""` the Name shows as `" "`, and `""` / `"Lovelace"` gives `" Lovelace"`.
- CSV output gets a literal `N/A` for missing fields, so scripts can't tell it apart from a real value. `format_value` writes an empty cell for null instead.
- A value that's literally `"N/A"` is treated as missing (unlikely, but it follows from the design).

Returning `Option<&str>` and only picking the fallback at output time (`unwrap_or("N/A")` for the table, `unwrap_or("")` for CSV) fixes all three and removes the match:

```rust
fn field(&self, object: &str, key: &str) -> Option<&str> {
    self.value[object][key].as_str().filter(|s| !s.is_empty())
}

fn full_name(&self) -> Option<String> {
    let parts: Vec<&str> = [self.field("user", "first_name"), self.field("user", "last_name")]
        .into_iter()
        .flatten()
        .collect();
    (!parts.is_empty()).then(|| parts.join(" "))
}
```

It would also be worth adding a test with missing/empty names and no `workspace.url`. The current fixture is fully populated, so none of these paths run.

sergey-borovkov (commented):
A few small things inline, nothing blocking.

src/commands/whoami.rs
```diff
@@ -0,0 +1,186 @@
...
+                wtr.write_record([
+                    "email",
+                    "name",
+                    "user_id",
+                    "workspace",
+                    "workspace_id",
+                    "workspace_url",
+                ])
+                .unwrap();
+                wtr.write_record([
```
sergey-borovkov: Nit: this is the same six fields as the table branch above, listed a second time. Adding or reordering a field means keeping both lists in sync. A single `rows()` returning `(label, csv_header, value)` tuples that both branches iterate over would avoid that.

src/commands/whoami.rs
```diff
@@ -0,0 +1,186 @@
...
+                "first_name": "Ada",
+                "last_name": "Lovelace",
+                "email": "ada@example.com"
+            }
+        })
+    }
+
+    #[test]
+    fn test_whoami_get_returns_profile() {
+        let _tmp_dir = tempdir().unwrap();
```
sergey-borovkov: Nit: `_tmp_dir` isn't used, since nothing overrides HOME or writes to it. This line and the `tempfile` import can go.

src/commands/whoami.rs
salmanfarisvp: Good catch, "`N/A`" shouldn’t double as the missing marker. I’ll switch field/full_name to Option, use "`N/A`" only for the table and `" "` for CSV, and add a test for empty/missing names and no workspace.url.

src/commands/whoami.rs
salmanfarisvp: Agreed, I’ll add a shared `rows()` returning `(label, csv_header, value)` so table and CSV stay in sync.

src/commands/whoami.rs
salmanfarisvp: That `tempdir` isn’t used. I’ll remove it and the `tempfile` import.

sergey-borovkov (approved):
All three comments addressed, plus the stdout fix in `commands::get`. Checked out the branch: 222 tests pass, clippy and `cargo fmt --check` are clean, and with a bad token stdout is now empty so `-o json` stays parseable.

